PR #42, opened 2016-01-25
New configurable CALIFA demonstrator geometry macro.

hapol:
The macro create_califa_demonstrator.C has been modified to allow the definition of CALIFA Demonstrator geometries. User can select the number of petals included, the type of petal (single, double) in each sector and a geometrical transformation per petal. A simple set of parameters is used for inserting the petals and move them, including a short howto about their use.
The macro create_califa_s438b_geo.C has been deleted: it was a test which did not correspond to the s438b real geometry. S348b geometry is now included as an example in create_califa_demonstrator.C


hapol:
Sorry, forget about this one, I saved it before changing to R3BRootGroup:dev.

PR #65, opened 2016-03-22
NeuLAND Update 1603

janmayer:
Save an image of the created geometry

Use LandPoint if NeulandPoints not available

Macros for different NeuLAND detector sizes

Updated macros

Gitignore files


kresan:
Please use for future pull requests a separated feature branch in your repo, and not the dev. Feature branches can be later deleted - after they have been merged with master.

PR #95, opened 2016-08-05
Feature pspx

isyndikus:
Update for PSPX, including parameter files for all 3 levels (MAPPED, CAL, HIT).


kresan:
merged with commits 49f2aef and c5615d3

PR #115, opened 2017-01-30
Feature pspx

isyndikus:
Small changes for PSPX


kresan:
merged with commit 0c9cb0f

PR #131, opened 2017-04-05
Don't force std=c++11 flag in neuland. One: It is 2017 already. Two: …

bl0x:
Don't force std=c++11 flag in neuland. One: It is 2017 already. Two: Setting something else (i.e. std=c++1y) in cmake gets overwritten.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,7 @@ If(NOT _HAS_CXX11_FLAG)
 EndIf()
 
 Execute_process(COMMAND $ENV{SIMPATH}/bin/fairsoft-config --cxxflags OUTPUT_VARIABLE _res_fairsoft_config OUTPUT_STRIP_TRAILING_WHITESPACE)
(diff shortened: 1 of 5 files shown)


kresan:
We will stay with c++11 until there will be no need in supporting older gcc versions.
In general, @janmayer could you please minimise usage of locally defined compiler flags in neuland sub-module? Preferably we have to solve it on the level of main CMakeLists.txt, and not have modularized settings.

bl0x:
This wasn't pulled so far... why not? The check in the main CMakeLists.txt throws off newer compilers which set newer flags by default in fairsoft-config. In may case fairsoft-config --cxxflags returns --std=c++1y. The check in the main CMakefile fails and I cannot compile without making the changes in this commit.

kresan:
@bl0x should I close this one? Is included in #138

bl0x:
No, forget about this one.BastiiAm 19.05.2017 08:01 schrieb Dmytro Kresan <notifications@github.com>:@bl0x should I close this one? Is included in #138

PR #135, opened 2017-04-17
feature_PSPX: correct Doxygen documentation

isyndikus:


kresan:
merged into release candidate branch with commit c08d389
will be synced with dev automatically later

PR #137, opened 2017-05-15
Renaming Califa Classes

PabloCabanelas:


kresan:
Great job! Still some minor remarks before merging:

- class R3BCalifaCrystalCal2HitPar is the parameter container for HIT level, so the proper name would be R3BCalifaHitPar. CrystalCal2HitPar is appropriate for the task, which creates HitPar containers.

- Maybe we can find a better name for R3BCallifaPIDCalFiller. This Filler suffix we do not have anywhere else. What is this task doing?

- The rebasing did go properly. You discarded the latest changes in following files:
create_aladin_geo.C
create_cal_geo.C
create_mfi_geo.C
create_mtof_geo.C
create_tof_geo.C
Please take these files from central dev, commit to your branch, and then do the squashing. Files should disappear from the squashed commit.

You can leave this pull request open, push your branch with modifications and notify here.

Thanks!

PabloCabanelas:
Points 1 and 3: sure! It was my mistake, sorry. I do not know why the rebase didn't go properly... :-/
Point 2:  Ok, got it. I think we can find a more suitable name for the class.

I let you know here once it is fixed.

Thanks!

PabloCabanelas:
Branch is fixed. Should be ready to merge. Thanks!

PR #138, opened 2017-05-17
Feature whiterabbit reader

bl0x:
Please pull soon! Oleg needs this.

### DIFF
--- a/neuland/CMakeLists.txt
+++ b/neuland/CMakeLists.txt
@@ -28,7 +28,7 @@ set(DEPENDENCIES
 )
 
 # -pthread for gtest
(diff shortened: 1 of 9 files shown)


kresan:
The change in the main CMakeLists.txt breaks the compilation. Now you explicitly require that FairSoft was compiled with std=c++1y. This is not the case, for example, on GSI Linux cluster.

We need to modify the check of FairSoft compiler flags in the way: error is thrown if both c++11 or c++1y were not used.

If Oleg needs the time stamp reader urgently, please split the commits.

bl0x:
I've reverted and pushed again.

PR #139, opened 2017-05-19
Check also for c++1y flag in fairsoft-config.

kresan:


kresan:
@bl0x can you check if it works for you?

PR #146, opened 2017-06-22
Glad magnet position fix backcompatibility

miharenthas:
As Dima is already aware, I had a minor issue with GLAD positioning: the 14 degrees rotation on the -Y axis was not applied when GLAD was constructed with the "just-by-name" constructor, leading to a buggy behaviour of code written before commit f3eb7099988ca632cfdd6c83a3aacbb659b4709a.

This issue turned out to be due to an unitialized variable used to define the rotation --which later would overwrite the one read from the geometry file.

The proposed fix in this branch (a little more formal than suggested on the forum) uses four #define'ed macros to pass the placement information to the parent class' constructor.
The scrips "r3bsim.C" and "run_sim.C", in their latest version, seems to run without errors and GLAD's placement seems to be all right.

Cheers,
Lorenzo Zanetti.


kresan:
merged with commit 502722a

for the next time please use R3BRootGroup:dev as base fork (instead of master) during creation of a pull request. Thanks.

PR #148, opened 2017-07-07
Fixes for Root5 and landreco test

janmayer:
Fixes for failing build under Root 5 (sorry).
landreco test might still fail, not sure what to do about that.

### DIFF
--- a/macros/r3b/r3ball.C
+++ b/macros/r3b/r3ball.C
@@ -211,6 +211,10 @@ void r3ball(Int_t nEvents = 1,
     run->AddModule(new ELILuMon(((TObjString*)fDetList->GetValue("LUMON"))->GetString()));
   }
 
+  // DEPRECATED: NeuLand Scintillator Detector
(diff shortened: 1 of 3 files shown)


kresan:
Thanks for fixing.
Previously today I have made "landreco" test running by changing SCINTNEULAND to LAND in the macros/r3b/land/r3blandsim.C Both names are not ok, so let's fix the convention later.

!IMPORTANT! To avoid extra efforts in supporting two ROOT versions in the future, and to allow faster development with better tools, we will drop ROOT 5 support from now on.
It will not appear in the next FairSoft (most probably aliBuild/alfadist release), and if somebody wants to use it, there are stable versions R3BRoot apr17 and FairSoft mar17.

Should I merge this commit anyhow?

janmayer:
The changes are quite small, so I think its fine to merge. Once Root 5 support has been dropped officially, (e.g. from travis/cdash and with check in the main CMakeLists), we can focus on making the code better.

PR #155, opened 2017-09-01
Use only Reference Physics List

kresan:
- Remove obsolete plists directory - use only reference physiscs list.
- Remove obsolete macros in gconfig directory.

Note: this clean-up is not yet finished. To be continued.


janmayer:
Wohooo! 👍

PR #185, opened 2018-02-01
Config files

miharenthas:
Just for fun: I found that for some applications I'd like to have another variable, pointing to R3BRoot --I called it R3BROOTPATH and added to the config.{csh,sh}.in.
Also, and that's a more serious request, I edited the config file templates so that system paths are not rewritten but appended --rewriting them could lead to breakages of other applications, and thus it's not a great idea in the first place, IMHO.

To allow for R3BRoot addition, I also copied and slightly modified the module WriteConfigFile.cmake from FairRoot --it's largely the same and this might be not the best way to do it, just the cheapest in term of time: suggestions are very welcome (especially since I'm not really familiar with cmake).


miharenthas:
I updated the config file templates to reflect those in FairRoot:dev (just merged): now we also have the command line options --in general, it's the same template with a variable added. Again: there's a lot of redundancy here and I'm very open to suggestions on how to make this leaner.

PR #187, opened 2018-02-13
FairFileSource

kresan:
- Switch from SetInputFile to SetSource (using FairFileSource).
- Proper handling of friend files using FairFileSource.


janmayer:
Well, I sure hope there is a good reason to make something as essential as setting an input file more annoying. This also breaks the symmetry of `run.SetInputFile("in.root"); run.SetOutputFile("out.root");`.

I've already created big functions just to set the parameter files for the runtimeDB, and with this I might as well go and write a full wrapper around FairRunAna and FairRunSim...

PR #191, opened 2018-02-20
R3BPspxCal2Hit.h: added inclusion of math.h (same as cmath) to suppor…

miharenthas:
Hi,
I got an inclusion miss during the last build. Apparently, math.h was included implicitly somewhere and now it isn't any more. Just added it --it's needed for isnan() checks.


vadimr3b:
Please put the include into the .cxx.

miharenthas:
Moved.

Issue #215, opened 2018-09-25
"Fatal" is deprecated


kresan:
Should be replaced by LOG('level')

kresan:
resolved with #224

PR #224, opened 2018-11-06
Replace obsolete Fatal() with LOG(fatal).

kresan:


kresan:
resolves #215

janmayer:
I'm pretty sure `FATAL` needs to be uppercase (https://cdash.gsi.de/viewBuildError.php?buildid=191422, https://github.com/FairRootGroup/FairRoot/blob/dev/fairtools/FairLogger.h)

For the new versions where the logger is extracted to FairSoft, linking requires `libFairLogger.so` in `FairSoft/lib`, which by chance is the same as `${Boost_LIBRARY_DIRS}`. Maybe adding it to `link_directories` in the main `CMakeLists.txt` will work.

I don't think FairRoot v18 ist required quite yet, even though a better / up to date solution for travis would be nice.

janmayer:
Ah, I see.

I would like to do some changes which would result in avoidable rebase-merge-headaches, most notably moving some code from `land/` to `neuland/` and a proposal for enforcing code formatting of changes files via travis. 
Should I go ahead with it or wait for completion of the two open PRs?

kresan:
The plan is to update the master branch after the current 2 PRs. We will then have stable master and can proceed to your proposal. I expect it to happen very soon.

PR #238, opened 2019-02-13
AMS: Print parameters for the DAQ

jose-luis-rs:
AMS: Print parameters for the DAQ


kresan:
please rebase and solve conflicts

kresan:
merge is not the right solution in this case. use rebase

PR #242, opened 2019-02-19
NeuLAND Preexp with EPICS

janmayer:
Co-authored-by: Hans Toshihide Törnqvist<hanstt@users.noreply.github.com>


janmayer:
Igor and @hanstt created these classes for pre-experimental configuration of NeuLAND. I have extracted them into a themed commit, moved them to `neuland/`, rebased to the current dev, and ran clang and cmake format. 

@kresan This does extend the responsibilities & dependencies of R3BRoot, which might not be desired. However the classes do use R3BRoot functionality and would be beneficial to have this functionality bundled in.

kresan:
I have no objections

PR #261, opened 2019-07-30
R3B distribution

vadimr3b:
I added some abstraction for random number distributions.

This should improve the usability, since it removes the necessity to edit the source code to change for example the beam energy distribution (see changes in R3BPhaseSpaceGenerator).

In addition this should improve the readability of the source code, since it removes the mathematical and technical details (see changes in R3BCosmicGenerator).

Please let me know, if anybody has some suggestions, on how to improve this PR.


vadimr3b:
PR#18 in the macro repo will fix the failing test.

janmayer:
Hey, good idea working on the generators. I always felt my sense of order tingle looking at the Cosmic Generator code that someone wrote in Fortran decades ago.

Here are my thoughts in descending order of importance:
- R3BDouble
    My first reaction was "Wait what?", then "WTF?", followed by "Why would you do that?". One should be able to safe primitive values in the Tree, right? R3BDouble doesn't even derive from TObject. And then using shared pointers with them seems like extraordinary overkill.
- R3BDistribution
    - While I like the general idea and you probably had fun building them, I'm not fully convinced that it was actually necessary. I wonder if there are existing libraries that do most of what you need? Either in `<random>`, somewhere in Boost, or similar (I didn't check)?
    - Why do `fValues` (+`CreateValues`) and `GetValueAddresses` even exist? Super complicated. There is probably a better way to store them in a tree.
    - Why does the user have to supply random values? `Array GetRandomValues(const Array rnd)` That is *exactly* what the class should be handling on its own. No need to bother the user with that.
    - There are no tests for the distribution classes.
    - The distribution classes do not really belong in `r3bgen/` - what about moving them to a new `utility/` folder or similar?

- General
    - When implementing derived classes, use the `override` keyword, e.g. in `R3BCosmicGenerator`
    - Imho, no test required to build & test R3BRoot should be in the external macro directory
    - The files are not formatted with clang-format
    - Consider deleting unnecessary comments, e.g. the header in `R3BCosmicGenerator.cxx` and the spacer comments (`// ---------`)

Have to tried to reproduce distributions like in the Sn-Test-Files with the PhaseSpaceGenerator? I had trouble with that and never got to fully fixing it.

vadimr3b:
Hi Jan,

thank you for your comments:

> R3BDouble
> My first reaction was "Wait what?", then "WTF?", followed by "Why would you do that?". One should be able to safe primitive values in the Tree, right? R3BDouble doesn't even derive from TObject. And then using shared pointers with them seems like extraordinary overkill.

You can save prim values in a Tree, unfortunately, but FairRoot does not provide an interface therefore (or maybe I just could not find it?).

> While I like the general idea and you probably had fun building them, I'm not fully convinced that it was actually necessary. I wonder if there are existing libraries that do most of what you need? Either in <random>, somewhere in Boost, or similar (I didn't check)?

I could not find anything. If there is something, we could exchange it.

> Why do fValues (+CreateValues) and GetValueAddresses even exist? Super complicated. There is probably a better way to store them in a tree.

I agree, but due to missing FairRoot interface, this was the only solution I could think of.
The best solution would be, to have an interface to store plain numbers. 

> Why does the user have to supply random values? Array GetRandomValues(const Array rnd) That is exactly what the class should be handling on its own. No need to bother the user with that.

I would disagree here. I decide to go this way, so its actually more simple if you want to be able to pass the seed. This way, you can pass the seed you want to the generator and do not have to bother passing the desired seeds to the distributions. I think it also separates the tasks better: the distributions only reflect the distributions ;).

> Consider deleting unnecessary comments, e.g. the header in R3BCosmicGenerator.cxx and the spacer comments (// ---------)

Done.

> When implementing derived classes, use the override keyword, e.g. in R3BCosmicGenerator

I only wanted to adapt to the new feature and did not pay attention to it. Apparently several definitions in r3bgen do not use the `override` specifier.

> Imho, no test required to build & test R3BRoot should be in the external macro directory

I fully agree. I removed the old ties.

janmayer:
> You can save prim values in a Tree, unfortunately, but FairRoot does not provide an interface therefore (or maybe I just could not find it?).

Hm, what values in which context do you want to save, anyway? Maybe there is a class (with a better name and some context) hidden somewhere that we can give life to and save to the tree instead.
I do find the `R3BDouble` and the pointer-ing around it quite nasty,

> I would disagree here. I decide to go this way, so its actually more simple if you want to be able to pass the seed. This way, you can pass the seed you want to the generator and do not have to bother passing the desired seeds to the distributions. I think it also separates the tasks better: the distributions only reflect the distributions ;).

OK, i understand. The classes take a uniform distribution and shape it to something non-uniform.

vadimr3b:
> Hm, what values in which context do you want to save, anyway?
My idea was to be able to get the random values in case you want to do further calculations with it.
For example: If you want to set the beam energy depending on the z position of your vertex (i.e. taking energy loss into account)

> Maybe there is a class (with a better name and some context) hidden somewhere that we can give life to and save to the tree instead. I do find the R3BDouble and the pointer-ing around it quite nasty,
If there is a better solution we should of course stick to it.

PR #264, opened 2019-08-29
Remove constexpr declaration.

kresan:
In c++ 11 sin() is declared as non-constexpr. Therefore make Sqr() and Sin2() functions non-constexpr.


vadimr3b:
Apparently sin is not a constexpr, therefore Sin2 should not be declated as constexpr.
Sqr on the other hand is a constexpr and can stay like this.

PR #273, opened 2019-10-03
Added name for the new AMS geometry that will be used for S455 and S4…

jose-luis-rs:
…67 experiments


kresan:
merged with 92c7680

PR #278, opened 2019-10-09
febex energy, nf, ns now signed.

klenze:


kresan:
please rebase from latest dev

kresan:
merged with 2b83417

PR #282, opened 2019-10-17
Califa Cal, Hit fixes

klenze:


hapol:
I'm going to push soon a new version of the CALIFA simulation that makes many changes in this pull request invalid or outdated. The present simulation is faulty by many reasons and the new version contains the full BARREL+iPhos, solving the found problems. It's now under review.
If you prefer to integrate this pull request now, I will rebase on top of it, and check if all your rewrite can still work.

I would request your revision of the Nf, Ns treatment for the new simulation version.

Some changes (FEBEX energy) and other corrections not related with simulation could be applied now.

kresan:
2 tests connected to CALIFA are failing

klenze:
I debugged some of the test problems. (Turns out that the problems I had locally while testing were harder than the ones found by travis. That will teach me to run tests locally instead of reading the cdash output.) 
I also had to edit some (test case) macros, so unless travis is clever enough to test with my macros PR, it will probably still fail, sorry. Let the record show that it was not /my/ decision to split off the test cases into a separate repository. 
Héctor, most of the work I did was independent of the geometry, it just assumed that there are crystal IDs, and that R3BCalifaGeometry can give the position for a given ID. The work on R3BCalifaGeometry is rather simple, I changed the GetAngle method to return a TVector3 ref, cached the requests in a std::map (so one does not need to build strings and do a directory lookup for every hit) and returned a TVector3 of nan's in case of error. 
Also, please note that I am still working on the spring 2019 data (esp the AmBe source run), so I would ask you to preserve some way to handle that geometry. (I am not particularly attached to the present ucesb mapping, so change that however you want.) I mean, we even still have test cases for Aladin active :-)
Good night,
   Philipp

kresan:
Can not accept this PR
1. Compilation breaks.
2. Branch is not properly rebased.
3. I do not see misc directory as part of R3BRoot.

janmayer:
Ah, I see you find TClonesArrays inappropriate as well. 

Might I interest you in my [TCAConnector](https://github.com/R3BRootGroup/R3BRoot/blob/f29a2a15ce48727ca8b58a2962738fb1e82f2bda/neuland/shared/TCAConnector.h)? Example usage: [Header](https://github.com/R3BRootGroup/R3BRoot/blob/f29a2a15ce48727ca8b58a2962738fb1e82f2bda/neuland/simulation/R3BNeulandPrimaryClusterFinder.h) [Code](https://github.com/R3BRootGroup/R3BRoot/blob/f29a2a15ce48727ca8b58a2962738fb1e82f2bda/neuland/simulation/R3BNeulandPrimaryClusterFinder.cxx)

hapol:
The new version modifies not only the geometry. The simulation is now faulty because is using volume iNode for the interaction classification, which is not related directly to the crystal, and this is being corrected. A new scheme based on CalifaPoints is also introduced, with a CalifaDigitizer that produces CalifaCalData with the proper crystal resolution (it was not the case before), given the possibility of a more detailed simulation response, needed for detailed studies (resolution dependence on crystal, ...). There are more differences that apply to the complete simulation procedure.
Said that, I agree that your suggestion of a map for the angles is fine and should be applied.
Regarding the preservation of the geometry: I would like to move to a simple and unique geometry from now on, which reduces dramatically the code and the complexity. To cover 2019 spring data, I would suggest that the geometry of the experiment is based on the new geometry. They require the construction of the geometry file and the modification of the lookup table (only the crystalID column). The rest of the code will be insensitive, if I'm not wrong.

klenze:
Sorry for the last PR. The new version is rebased to dev, misc is removed. Travis build succeeds, but it still fails test [califasim2](https://cdash.gsi.de/testDetails.php?test=5927227&build=217665). One error is the call to R3BCalifaCrystalCal2Hit::SetAngularWindow(), which is removed in my macros PR. 
The other error one (allocation of incomplete type 'FairParRootFileIo') I encountered yesterday locally and I think I 'fixed' it by running the macro with the + option, which should also be included in the macros PR. 

Jan, obviously I find raw TCAs inappropriate. I think our approaches are similar: wrapping the TCA in a type-templated class passing something which plays nice with range-based for loops. You used std::vector, I [wrapped](https://github.com/klenze/R3BRoot/blob/cal_fixes/califa/ROOT_template_hacks.h) the TIter class. Iterators might be a bit more light-weight than vectors, but on the other hand my version does not handle inserts. BTW, I would recommend replacing your C-style static casts (e.g. (T*)fTCA->At(i)) with dynamic_casts. I think the minimal overhead is well worth the additional type safety. 
(Also, I learned that my g++ 4.8.2 does not really now about c++11's std::runtime_error(const char*) and thus fails to compile your code. I used an ugly preprocessor macro to cast it explicitly (one can cast TString to const char*, and const char* to std::string), but at least it motivated me to work on compiling fairsoft with clang9 again. If that fails, I might PR the preprocessor hack with appropriate guards on the compiler version). 

Hector, ok, sounds good.

kresan:
Test is now successful (after merging PR in macros).
From my side all fine.
@hapol do you approve?

PR #285, opened 2019-10-23
Adapt to upcoming cmake changes in FairRoot

rbx:
This is just for your reference.
But it should also be safe to merge, since it is all backwards-compatible.


kresan:
@rbx thanks a lot! Will keep it until the next FairRoot release

rbx:
Removed the `set(PROJECT_LIBRARY_PROPERTIES ${FAIRROOT_LIBRARY_PROPERTIES})`, we now solve it on FairRoot side.

kresan:
merged with 6ce2a21 and 53a3023

PR #287, opened 2019-10-24
sprintf bug fix

klenze:
travis compiler said: note: 'sprintf' output between 2 and 12 bytes into a destination of size 1. 
translation: random bad stuff *will* happen. 

Regarding sprintf, see http://natashenka.ca/wp-content/uploads/2014/01/sprintf8x11.png

The other 769 uses of sprintf may or may not be safe.

### DIFF
--- a/fiber/R3BBunchedFiberCal2Hit.cxx
+++ b/fiber/R3BBunchedFiberCal2Hit.cxx
@@ -106,8 +106,8 @@ InitStatus R3BBunchedFiberCal2Hit::Init()
 
   // ToT SAPMT:  
   for(Int_t i=0;i<4;i++){
-    char number[1];
-    sprintf(number, "%d", i);
+    char number[15];
+    snprintf(number, sizeof(number), "%d", i);
     chistName=fName+"_ToT_SAPMT"+number;
     chistTitle=fName+" ToT of single PMTs "+number;
     fh_ToT_s_Fib[i] = new TH2F(chistName.Data(), chistTitle.Data(), 2100, 0., 2100., 200, 0., 200.);   	   


janmayer:
> Regarding sprintf, see http://natashenka.ca/wp-content/uploads/2014/01/sprintf8x11.png

He said, and proceeded to use something almost as bad. 
Have you heard about our lord and savior, the C++ standard library?
https://en.cppreference.com/w/cpp/string/basic_string/to_string

klenze:
> Have you heard about our lord and savior, the C++ standard library?

"First, they came for the C strings and I said nothing, because I did not use C strings. 
Then they came for the C arrays and I said nothing because I used STL containers. 
Then they came for the function pointers and I said nothing, because I used std::function. 
Then they came for cascaded lambda expressions, and no one was left to speak for me."
(With apologies to Niemoeller.)

Realistically, getting rid of C string handling is probably not possible in this project, not while keeping [ROOT](https://github.com/root-project/root/blob/master/hist/histpainter/src/THistPainter.cxx#L3961) anyhow. Anyhow, I find format strings useful [sometimes](http://yosefk.com/c++fqa/io.html#fqa-15.1):

> Assignment Number 1 - convert all your evil printf("0x%08xn", x) statements to this:
> `std::cout << std::hex << std::setfill('0') << std::setw(8) << x << std::dec << std::endl; `

(I tried boost::format, but did not like it for some reason.)

The only reason for this PR was the fact that the compiler correctly noted that the code will lead to undefined behavior -- I will certainly not try to fix style choices I might dislike (Like [monster switch statements](https://github.com/R3BRootGroup/R3BRoot/blob/dev/evtvis/R3BIonName.cxx#L27) instead of array/map lookup). I chose snprintf to make people aware of the existence of a safe alternative to the stuff they use. For functionality, getting rid of number and instead using `TString operator+(const TString &s, Long_t i)`, e.g.
`chistName=fName+"_ToT_SAPMT"+(long)i;`
would probably have been the cleanest solution here.

janmayer:
> "First, they came for the C strings and I said nothing, because I did not use C strings.
> Then they came for the C arrays and I said nothing because I used STL containers.
> Then they came for the function pointers and I said nothing, because I used std::function.
> Then they came for cascaded lambda expressions, and no one was left to speak for me."

"Without a Standard Model, there is naught but madness and Chaos among the stars." -Aetheric Manuscripts Appendix 71


> Realistically, getting rid of C string handling is probably not possible in this project

"The Emperor extends His will to us when we destroy the wicked and impure to reconquer the galaxy in His name. The Emperor leads the galaxy to righteousness, and thus we must petition for His judgement on the wicked... for it is judgement without mercy." -High Marshal Helbrecht of the Black Templars


> The only reason for this PR was the fact that the compiler correctly noted that the code will lead to 
undefined behavior

"He who prosecutes undefined behavior with zeal, is the ally of Mankind." -Gathalamorians 60.41


>I will certainly not try to fix style choices I might dislike

"Watch your brother, for his sin of heresy is thy sin of tolerance." -Macharians 16.4


>And I find format strings useful sometimes


"From such small heresies are great apostasies made." -Maxims Metalica, 76.12

> I tried boost::format, but did not like it for some reason.

"Do not ask the Omnissiah for a pleasant existence, ask Him for the strength to overcome our tribulations." -Aphorisms 62 (https://en.cppreference.com/w/cpp/utility/format)

> ... to make people aware of the existence of a safe alternative to the stuff they use

"Let the weaker in mind follow the words of the strong." -Canticles of Gethsemenoth 87

(/ooc With apologies to the [Adeptus Mechanicus](https://wh40k.lexicanum.com/wiki/Adeptus_Mechanicus). I had way to much fun with [this](https://www.reddit.com/r/AdeptusMechanicus/comments/asq45q/).)

kresan:
merged with 777ed4f

PR #291, opened 2019-10-29
Travis clang format

janmayer:


janmayer:
Ok, should be ready for review now:

This PR changes two things with the Travis-CI checks:

1) `clang-format` checks for changed files: 
    - If a file was touched, `clang-format-8` will be applied to it on travis. 
    - If there are changes, the job will show as failed and the problems are displayed [like so]( https://travis-ci.org/R3BRootGroup/R3BRoot/jobs/604853675)

    I think this is the least intrusive way to slowly get the whole code base formatted.
    A statically build binary of `clang-format` (version 8) that should run on any Linux is available [here](https://github.com/angular/clang-format/raw/master/bin/linux_x64/clang-format) for those without the ability to install stuff - no excuses.

2) Refactored `.travis.yml`:
     - Use parallel jobs
     - Removed all language specific settings and set the language to minimal (build/tests happen in docker anyway)
     - Unzipping the docker image is not necessary (saves a few minutes)
     - Extracted generation of `Dockerfile` to its own script

hapol:
>clang-format --version
clang-format version 9.0.0 (tags/google/stable/2019-05-14)

Any problem with this version? Your comments explicitly mention version 8...

janmayer:
> > clang-format --version
> > clang-format version 9.0.0 (tags/google/stable/2019-05-14)
> 
> Any problem with this version? Your comments explicitly mention version 8...

It should be fine™. 
The reason for the version requirement is that the only statically build version I could find (linked above) is 8.0.0 - so people can just fetch that one binary and it works. I am open for better solutions 😄

PR #299, opened 2019-11-29
Add GPL 3 License

kresan:


janmayer:
That's the big PR I've been waiting for. 13 LoC added to each file - not good, not terrible. 

You might consider adding  `// clang-format off` to all LinkDef files where it is not present already such that the `+`es are not removed from the class name (not sure if that matters).

PR #310, opened 2020-01-16
Added -Werror, (minus some). Fixed warnings. 

klenze:
While I welcome enforcing formatting rules, I believe we can do even better. 

To that end, I added the gcc -Werror flag (minus sign-compare, reorder, unused*) to the default flags.

It can be disabled by setting export NO_WERROR=1 in the shell, so it should not break icc and the like. 

Also I took care of the warnings. 
Some of them clearly indicated bugs (e.g. for (Int_t i; i<4; i++), some are mostly harmless (a #pragma link C++ where gcc can see it). 

Potential use of undefined variable was a very common theme. 

Typically:

double d;
for (...)
   if (...)
      d=...;
For doubles, I generally initialized them nan. For non-negative integers, I initialized them to -1 and mostly assert()ed after the loop that they were assigned. 

Cheers,
   Philipp

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,15 @@ find_package(HEPMC)
 # Check the compiler and set the compile and link flags
 check_compiler()
 
-#set(CMAKE_CXX_FLAGS -std=c++14)
+if(DEFINED $ENV{WERROR} AND $ENV{WERROR})
+  message(STATUS "Will compile with -Werror. ")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wno-error=sign-compare -Wno-error=reorder -Wno-error=unused-variable -Wno-error=unused-but-set-variable")
+else()
+  if (NOT APPLE)
+    message(STATUS "Set env WERROR to 1 to enable -Werror. If origin/dev compiles on your platform with that option, it is definitly a good idea to do that.")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-sign-compare -Wno-reorder -Wno-unused-variable -Wno-unused-but-set-variable")
+  endif()
+endif()
 
 set(LIBRARY_OUTPUT_PATH "${CMAKE_BINARY_DIR}/lib")
 set(EXECUTABLE_OUTPUT_PATH "${CMAKE_BINARY_DIR}/bin")
@@ -217,6 +225,9 @@ find_package(ucesb)
 if(ucesb_FOUND)
     set(WITH_UCESB true)
     set(LD_LIBRARY_PATH ${ucesb_LIBRARY_DIR} ${LD_LIBRARY_PATH})
+else()
+  message(WARNING "${BMagenta}No UCESB configured. You will not be able to unpack lmd files. Set UCESB_DIR"
+    " to a ucesb installation to fix this.${CR}")
 endif(ucesb_FOUND)
 
 find_package(EPICS)
--- a/R3BRoot_test.cmake
+++ b/R3BRoot_test.cmake
@@ -8,7 +8,8 @@ Set(CTEST_PROJECT_NAME "R3BRoot")
 Find_Program(CTEST_GIT_COMMAND NAMES git)
 Set(CTEST_UPDATE_COMMAND "${CTEST_GIT_COMMAND}")
 
-Set(BUILD_COMMAND "make")
+Set(BUILD_COMMAND "make -k")
+
 Set(CTEST_BUILD_COMMAND "${BUILD_COMMAND} -j$ENV{number_of_processors}")
 
 Set(EXTRA_FLAGS $ENV{EXTRA_FLAGS})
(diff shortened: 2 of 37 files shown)


kresan:
Very nice PR. Please fix compiler error(s)

kresan:
merged with 56170ac

CMakeLists.txt
```diff
@@ -123,7 +123,10 @@ find_package(HEPMC)
 # Check the compiler and set the compile and link flags
 check_compiler()
 
-#set(CMAKE_CXX_FLAGS -std=c++14)
+if(NOT DEFINED ENV{NO_WERROR})
```
kresan: Switching -Wall and -Werror on by default causes many problems on different platforms. I propose to have it optional and off by default

klenze:
@kresan, in various files, I have changed
`        tpatvalue = (itpat && (1 << fTpat_bit)) >> fTpat_bit;`
to
`        tpatvalue = (itpat & (1 << fTpat_bit)) >> fTpat_bit;`
This is a change in functionality. I did this because the original code did not seem to archive whatever it was trying to archive. 
(What I think the original code did was: For fTpat_bit zero, it sets tpatvalue to one iff itpat is nonzero. In all other cases, tpatvalue will be assigned to zero.)

kresan:
@klenze , then please open a new PR with the changes in functionality you have mentioned. In this state the current PR can not be accepted. You can not require such flag as -Werror without testing it on all platforms we currently support. It was breaking on GSI Linux (Debian 8) and macOS.

klenze:
@kresan, regarding the changes of tpatvalue, it seems that the original commit by Michael already used my preferred interpretation in a previous commit in some cases: 
`$ grep --exclude-dir=.git tpatvalue -r . | sed s/:.*// | sort | uniq | xargs -L 1 git blame -f | grep -E "tpatval.*>>" | sort
b6ae61e1 los/R3BLosCal2Hit.cxx (Dmytro Kresan         2019-11-29 12:44:43 +0100  457)         Int_t tpatvalue = (itpat & (1 << fTpat_bit)) >> fTpat_bit;
b6ae61e1 r3bbase/R3BOnlineSpectra.cxx (Dmytro Kresan         2019-11-29 12:44:43 +0100 1213)         tpatvalue = (itpat && (1 << fTpat_bit)) >> fTpat_bit;
c6c9d3c5 r3bbase/R3BGlobalAnalysisS454.cxx (Michael Heil  2019-08-20 11:09:10 +0200  555)             tpatvalue = (itpat && (1 << fTpat_bit)) >> fTpat_bit;
c6c9d3c5 r3bbase/R3BOnlineSpectraDec2019.cxx (Michael Heil  2019-08-20 11:09:10 +0200 1431)         tpatvalue = (itpat && (1 << fTpat_bit)) >> fTpat_bit;
c6c9d3c5 r3bbase/R3BOnlineSpectraS454.cxx (Michael Heil 2019-08-20 11:09:10 +0200 1147)         tpatvalue = (itpat && (1 << fTpat_bit)) >> fTpat_bit;
c6c9d3c5 tof/R3BTofdCal2Histo.cxx (Michael Heil 2019-08-20 11:09:10 +0200 207)         Int_t tpatvalue = (itpat & (1 << fTpat_bit)) >> fTpat_bit;
c6c9d3c5 tof/R3BTofdCal2Hit.cxx (Michael Heil          2019-08-20 11:09:10 +0200  342)         Int_t tpatvalue = (itpat & (1 << fTpat_bit)) >> fTpat_bit;`

I thus deem the change to be a bug fix and not a change in intended functionality. 
(The alternative would have been to change it to  itpat && (!fTpat_bit);, which is no functional change but would have stopped -Werror from complaining.)

I also left the Werror option in for travis, where all error-causing warnings are currently (hopefully) fixed. So when people add code, if they skipped the warnings locally, travis will complain. 

Cheers,
   Philipp

kresan:
I do not agree to having WERROR set for travis. The problem is following: people compile it locally - all fine, they send a PR, and in most of cases nobody looks at red failing cross and no action is taken. After a while people complain that the PR is still not merged. So, with what you propose, I will need to send a gentle reminder to everybody about fixing the warnings they have created.
To conclude: as agreed before - we are not yet at the stage to put such requirement by default. Thus it should also not be in travis.

klenze:
Okay. I still do not get the big difference between the handling of Werror and clang-format.
* Both would by default checked first checked on travis
* Both would lead to a PR not being merged, with a "travis check failed message" popping up.
* Both can be run run locally, provided relevant tools (clang-build, same g++ version as travis has) are installed. 

Personally, I feel people ignore compiler warnings at their own peril. There may be people among us who have never caused a segfault, but for mere mortals such as myself, fixing -Wuninitialized and friends in my code means fixing bugs most of the time. (There are also warnings about stuff which probably is not a bug, such as unused variables and the like. I did not turn these into errors, though.)

As a sort of compromise, I have disabled Werror for travis-generate-dockerfile.sh, as requested, but added a new wrapper script 
travis-generate-dockerfile-buildonly.sh, which enables Werror (and disables tests), which you could check for separately if you like, just as you do for the clang-format script. 

In other news, I will probably not try to add a clang-tidy check. I feel it might benefit the project, but I just don't have the spare time. 

Cheers,
   Philipp

CMakeLists.txt
```diff
@@ -123,7 +123,13 @@ find_package(HEPMC)
 # Check the compiler and set the compile and link flags
 check_compiler()
 
-#set(CMAKE_CXX_FLAGS -std=c++14)
+if(DEFINED ENV{WERROR} AND $ENV{WERROR})
```
kresan: Did not run on macOS. Did you forget ENV --> $ENV?

CMakeLists.txt
```diff
@@ -123,7 +123,13 @@ find_package(HEPMC)
...
+  message(STATUS "Will compile with -Werror. ")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Werror -Wno-error=sign-compare -Wno-error=reorder -Wno-error=unused-variable -Wno-error=unused-but-set-variable")
+else()
+  message(STATUS "Set env WERROR to 1 to enable -Werror. If origin/dev compiles on your platform with that option, it is definitly a good idea to do that.")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
```
kresan: Do you really want to see tons of warnings about hiding the overloaded virtual function and about order of members initialisation?? Absolutely disagree.

klenze:
Ok, I have removed the relevant warnings from Wall. Is it ok now?

kresan:
This is what I get if I build it on macOS:

`warning: unknown warning option '-Wno-unused-but-set-variable'; did you mean '-Wno-unused-const-variable'? [-Wunknown-warning-option]`

klenze:
@kresan, sorry. I do not have a apple build environment, and neither does travis, so testing against it is a bit hit and miss. I added an if (NOT APPLE), hope this does the job.

kresan:
@jose-luis-rs any objections against merging it now? Will it disturb running experiment?

kresan:
@klenze Tests are not executed on Travis and results are not published on CDash. Any idea why?

klenze:
> @klenze Tests are not executed on Travis and results are not published on CDash. Any idea why?

Yes, because I made a logic error, forgot a not in the tests cmake script. Should be fixed now.

PR #325, opened 2020-02-14
Deep Neural Network Module for NeuLAND; an ENSAR2 project by C. A. Douma

ChristiaanDoumaDNN:
Deep Neural Network Module for NeuLAND; an ENSAR2 project by C. A. Douma

Could these changes be added to the current dev branch?


janmayer:
Over SEVEN HUNDRED files, wait what? I expected just the streamlined Keras models - I'm still trying to wrap my head around all the stuff you did. Wow.  I think the important piece is [DNN/Keras/Script/NetworkDefinition.py](https://github.com/R3BRootGroup/R3BRoot/pull/325/files#diff-323b1e33d2c487c2b7e0dfec07dbe91f), right?

There are several copies of Digitizers, of Geometries, and other classes that already have a 
Its more its own separate project. And you even redefine Masses and Nuclei.

My recommendation is that you open a repository on github just with this code. I'm afraid that no one else will ever be able (or want) to maintain it.

... and why is there Fortran mixed in?

ChristiaanDoumaDNN:
The DNN itself is indeed defined in DNN/Keras/Script/ but this is by far
not the only important piece. For example, the TextFiles-folder is
important, as it
translates the ROOT-input to the DNN and also does the input normalization.
This step is extremely important and the DNN is quite sensitive to it.
Translator is also important, as it decided the 'correct' output for the
network training.

The DNN module is more than just some network definition, it is a
fully-automatized simulation and analysis method for NeuLAND, even with
its own UI to modify parameters (InputClass-folder).

The Fortran-piece is the so-called FOLD-code (used in C. A. Douma et al.,
EPJA 56 51 2020) used for DWBA theoretical predictions
of cross sections. I used it to compare the experimental g.s. cross
sections of the NeuLAND commissioning run at SAMURAI (2015) with theory.

However, as this DNN module does not conflict with any of the rest of
R3BRoot, I would still ask if it could be merged into the dev-branch
so that people who want to can take advantage of it.

Christiaan Douma.






Op vr 14 feb. 2020 om 16:26 schreef Jan Mayer <notifications@github.com>:

PR #328, opened 2020-02-19
working now with 2019/s444 geo file now again.

klenze:


klenze:
will fix it when our local machines are up again.

klenze:
The failed test is landreco. (relevant cdash item here, not that @travis-ci would care to link to it https://cdash.gsi.de/testDetails.php?test=6351475&build=228130 )

Let me state for the protocol that locally, the test works, and cdash gives very little indication of why it failed on travis. Unless I have more input (the full stdout/stderr of ctest -R landreco -VV would be a start), I can not debug this problem. 

Squashing commits for now. 
Cheers,
   Philipp

CMakeLists.txt
```diff
@@ -173,7 +173,9 @@ find_package2(PUBLIC Boost
 regex serialization log log_setup atomic date_time signals ${FairMQ_Boost_COMPONENTS}
 )
 if(Boost_FOUND)
-    set(Boost_Avail 1)
+  set(Boost_Avail 1)
+  message("Adding Boost_LIBRARY_DIRS=${Boost_LIBRARY_DIRS} to linker directories")
+  add_link_options(-L${Boost_LIBRARY_DIRS})
```
kresan: Why not to use Set(LINK_DIRECTORIES) in califa/CMakeLists.txt?

califa/CMakeLists.txt
```diff
@@ -83,3 +86,5 @@ endif(FAIRDB_FOUND)
 
 
 GENERATE_LIBRARY()
+target_link_libraries(R3BCalifa boost_regex)
```
kresan: boost_regex should go into set(DEPENDENCIES

califa/CMakeLists.txt
```diff
@@ -44,6 +45,7 @@ include_directories(SYSTEM ${SYSTEM_INCLUDE_DIRECTORIES})
 set(LINK_DIRECTORIES
 ${ROOT_LIBRARY_DIR}
 ${FAIRROOT_LIBRARY_DIR}
+${SIMPATH}/lib
```
kresan: Boost_LIBRARY_DIRS

califa/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@
 Set(SYSTEM_INCLUDE_DIRECTORIES
 ${SYSTEM_INCLUDE_DIRECTORIES}
 ${BASE_INCLUDE_DIRECTORIES}
+${SIMPATH}/include/
```
kresan: Boost_INCLUDE_DIR

klenze:
While editing the cmake file, I have also cleaned up a bit. I removed unused dependencies (xball), moved califa classes to subdirectories (no more confusion between tasks and calibration parameters) and added a readme file describing the classes.

califa/CMakeLists.txt
klenze: `[ 74%] Linking CXX shared library ../lib/libtimestitcher.so
825/usr/bin/ld: cannot find -lboost_regex
826timestitcher/CMakeFiles/timestitcher.dir/build.make:150: recipe for target 'lib/libtimestitcher.so.SOVERSION' failed
`
Note that this is not for califa, but for libtimestitcher.so. 
It seems dependencies get added transitively, while link directories don't.

klenze: the issue seems to be that target_link_libraries, as called by  $FAIRROOTPATH/share/fairbase/cmake/modules/ROOTMacros.cmake in GENERATE_LIBRARY, seems to add the ${DEPENDENCIES} (implicitly) PUBLIC, which will also add them to any user of that library. I do not think this is a good default behavior, but whatever. 

One way to solve this would be to add a PRIVATE in front of every dependency, another to fix ROOTMacros.cmake in FairRoot so all the stuff gets added private. 

The cmake-y thing to do, I feel, was to use ${Boost_REGEX_LIBRARY} (note the unique capitalization), which resolves to the full path of the relevant lib (plus possibly some [additional keywords](https://cmake.org/cmake/help/v3.15/module/FindBoost.html). 

Here is to hoping that it will compile on MacOS as well.

klenze:
@kresan: I can not seem to mark any more issues resolved.

See also notes on https://github.com/R3BRootGroup/R3BRoot/pull/328#discussion_r385585040

kresan:
it compiles on macOS. Please squash the commits.

Concerning the cmake targets in FairRoot. Well, this part is under heavy re-organization, which will affect R3BRoot at some point. Let's wait when it is done first on FairRoot side.

Issue #332, opened 2020-02-24
Source of truth for start time


janmayer:
For NeuLAND calibration and online analysis, we currently rely on other detectors. The problem is that these detectors change from experiment to experiment and within the experiment the offset can change (e.g. on DAQ reboot). 

For example, in the [current version of the NeuLAND Online Analysis Script]( https://github.com/R3BRootGroup/R3BRoot/pull/331/files#diff-7c26177db7ad9142911d9e5a59ef9228) we do this with an ifdef switch, taking LOS by default and  Sofia if it exists - not ideal. In addition, there is always some logic involved to extract the start time from the other detectors' data – something that does not belong in the NeuLAND code but in its own class
The offset is a parameter given in the online analysis script, 
```C++
    auto nlhit = new R3BNeulandCal2Hit();
    nlhit->SetDistanceToTarget(distanceToTarget);
    nlhit->SetGlobalTimeOffset(7300);
    run->AddTask(nlhit);
```
Which obviously makes analyzing runs with different offsets together impossible.

Is there a common class, a “source of truth”, that I can simply query/consume for the start time? That should also take care of the changing part of the time offset as a run parameter (ParFile).
Depending on the experiment, one would add the respective task for the right start detector (which fill the same Container with the start time) to the macro.
How do the other detectors handle the start time?

inkdot7:
In land02 this is handled by having an experiment-independent incoming tracking.  The available detectors are declared, together with positions (z) and their capabilities (x, y, dE, t).  Then the data is used to its fullest extent to calculate the position and time on target (using fits), as well as charge (through energy loss).  These sets of parameters are then available as x0, y0, t0 to other detectors.  One caveat is that the reconstruction of the incoming detectors needs to happen before other detectors that need this information runs.

Just a start detector is not enough, as it is not giving time-on-target - it needs extrapolation along z.  What land02 did not do very well was adapting to movement of detectors, since this info was in a C source file...

kresan:
Is the issue still relevant?

janmayer:
Providing the start time is something that affects the online and offline analysis of experimental data, which I only have limited insight into. As a NeuLAND developer, I just want to consume the start time.

I did "solve" it for some of the previous experiments in https://github.com/R3BRootGroup/R3BRoot/issues/332 but I’m interfering with the code of the other detector groups there.

[...]

PR #334, opened 2020-02-25
Provide TStart to R3BEventHeader

janmayer:
With respect to #332 :
- add TStart field to R3BEventHeader
- write Detector/Experiment specific Tasks that set that value, to be selected in macros
    - Los Example included
    - Sofia->Sofia Repo
- use value in NeuLAND Tasks

There is a TimeStamp field, but that is (rarely) used and I don't know its full meaning.


janmayer:
Sofia Time Provider: https://github.com/R3BRootGroup/sofia/pull/147

jose-luis-rs:
Sorry, I have seen your PR now. Please, rebase and push again for the acceptance.

PR #336, opened 2020-03-08
WIP: Adapt to new FairRoot CMake package

dennisklein:
@kresan This is very much work-in-progress. It needs a couple of discussions/decisions still to be completed that is why I opened the PR already. Let's have a talk this week or so.

I use R3BRoot to test and harden FairRootGroup/FairRoot#933, so changes there go hand-in-hand with changes here over the next few days (this PR depends on the FairRoot PR).


dennisklein:
Obsolete.

PR #348, opened 2020-05-04
Correct number of elements in loop after cal removal under threshold

hapol:


hapol:
Problem not solved... It will be reopened after correction.

PR #349, opened 2020-05-05
Support for the new version of ext_data_setup() in R3BUcesbSource.

bl0x:
Ucesb has gained the ability to get more detailed information about how the mapping went during the setup(). This introduces also a change in the interface, so quick action is needed.

Also fix up a comment that was outdated.


bl0x:
This change maintains backwards compatibility with older ucesb versions.
However, existing code may break with the new version of ucesb, if illegal mappings are performed which have so far not been discovered.

bl0x:
Please merge as you like.

kresan:
Please squash the commits

bl0x:
Oops, yes. Done.

PR #356, opened 2020-09-02
Added New Features For CALIFA Simulation

gabrigarjim:
Added New features for CALIFA simulation.
- Experimental parameters added to the simulation (thresholds and resolutions)
- New parameter container for simulation
- Fixed bug in the digitizer class


CMakeLists.txt
```diff
@@ -150,7 +150,7 @@ if(USE_PATH_INFO)
 endif(USE_PATH_INFO)
 
 # Check if the user wants to build the project in the source directory
-check_out_of_source_build()
+#check_out_of_source_build()
```
kresan, line 153: Why do you need to comment this out?

kresan:
You have 3 commits with exactly the same comment line. Please squash them into a single one

CMakeLists.txt
gabrigarjim: Hi,
it was giving an error while building in Mac: 
_Unknown CMake command "check_out_of_source_build"_
Is this command built in a recent version of FairRoot? (mine is v18.2.1)

kresan: On my Mac it works with FairRoot 18.2.1. It is defined here: https://github.com/FairRootGroup/FairRoot/blob/RC_v18.2.0/cmake/modules/FairMacros.cmake
So, something is wrong with your FairRoot installation.
This check is important and we can not just skip it.

gabrigarjim: Now its working, a re-installation solved the problem. Closing the PR and sending a proper one....

PR #362, opened 2020-10-26
Added a simple Generator for one single particle.

vadimr3b:
I do not know, if this is useful for anybody else, but I felt like there is a very simple single particle generator missing.
First, I wanted to simplify/update R3BIonGenerator, but I am not sure who uses it where.
That is why I added this class.

It uses R3BDistribution and allows a quick simulation setup:
`auto gen = new R3BParticleGenerator();`
`gen->SetProton(); `
`gen->SetEnergyDistribution(R3BDistribution1D::Flat(0, 1000));`
`gen->SetVertexDistribution(R3BDistribution3D::Delta( 0, 0, 0));`

Please let me know if you have any suggestions!

### DIFF
--- a/r3bgen/R3BBeamProperties.cxx
+++ b/r3bgen/R3BBeamProperties.cxx
@@ -1,0 +1,42 @@
+/******************************************************************************
+ *   Copyright (C) 2019 GSI Helmholtzzentrum für Schwerionenforschung GmbH    *
+ *   Copyright (C) 2019 Members of R3B Collaboration                          *
+ *                                                                            *
+ *             This software is distributed under the terms of the            *
+ *                 GNU General Public Licence (GPL) version 3,                *
+ *                    copied verbatim in the file "LICENSE".                  *
+ *                                                                            *
+ * In applying this license GSI does not waive the privileges and immunities  *
+ * granted to it by virtue of its status as an Intergovernmental Organization *
+ * or submit itself to any jurisdiction.                                      *
+ ******************************************************************************/
+
+#include "R3BBeamProperties.h"
+#include "R3BDistribution2D.h"
+#include "R3BDistribution3D.h"
+
+#include "TGeoPhysicalConstants.h" // for  amu_c2
+
+constexpr auto AMU_MeV = TGeoUnit::amu_c2 / TGeoUnit::MeV; // MeV
+
+R3BBeamProperties::R3BBeamProperties()
+    : fVertex_cm(R3BDistribution3D::Delta(0, 0, 0))
+    , fSpread_mRad(R3BDistribution2D::Delta(0, 0))
+    , fBeta(R3BDistribution1D::Delta(0.5))
+{
+}
+
+void R3BBeamProperties::SetEnergyDistribution(R3BDistribution<1> energyDistribution_AMeV,
+                                              int nNucleons,
+                                              double mass_MeV)
+{
+    const auto nucleonMass = (mass_MeV == 0. ? nNucleons * AMU_MeV : mass_MeV);
+
+    fBeta = R3BDistribution<1>([energyDistribution_AMeV, nucleonMass, nNucleons](
+                                   std::array<double, 1> values) mutable -> std::array<double, 1> {
+        const auto energy = energyDistribution_AMeV.GetRandomValues(values)[0];
+        const auto gamma = (nNucleons * energy) / nucleonMass + 1.;
+        const auto beta = std::sqrt(1. - 1. / (gamma * gamma));
+        return { beta };
+    });
+}
--- a/r3bgen/R3BBeamProperties.h
+++ b/r3bgen/R3BBeamProperties.h
@@ -1,0 +1,58 @@
+/******************************************************************************
+ *   Copyright (C) 2019 GSI Helmholtzzentrum für Schwerionenforschung GmbH    *
+ *   Copyright (C) 2019 Members of R3B Collaboration                          *
+ *                                                                            *
+ *             This software is distributed under the terms of the            *
+ *                 GNU General Public Licence (GPL) version 3,                *
+ *                    copied verbatim in the file "LICENSE".                  *
+ *                                                                            *
+ * In applying this license GSI does not waive the privileges and immunities  *
+ * granted to it by virtue of its status as an Intergovernmental Organization *
+ * or submit itself to any jurisdiction.                                      *
+ ******************************************************************************/
+
+#ifndef R3BBEAMPROPERTIES
+#define R3BBEAMPROPERTIES
+
+#include "R3BDistribution.h"
+#include "R3BDistribution1D.h"
+
+class R3BBeamProperties
+{
+  public:
+    R3BBeamProperties();
+
+    // Vertex distribution in cm
+    R3BDistribution<3>& GetVertexDistribution() { return fVertex_cm; }
+    void SetVertexDistribution(const R3BDistribution<3>& vertexDistribution_x_y_z_cm)
+    {
+        fVertex_cm = vertexDistribution_x_y_z_cm;
+    }
+
+    // spread in mRad
+    R3BDistribution<2>& GetSpreadDistribution() { return fSpread_mRad; }
+    void SetSpreadDistribution(const R3BDistribution<2>& spreadDistribution_theta_phi_mRad)
+    {
+        fSpread_mRad = spreadDistribution_theta_phi_mRad;
+    }
+    void SetSpreadDistribution(const R3BDistribution<1>& spreadDistribution_theta_mRad)
+    {
+        fSpread_mRad =
+            R3BDistribution<2>({ spreadDistribution_theta_mRad, R3BDistribution1D::Flat(0, 2 * TMath::Pi() * 1e3) });
+    }
+
+    // beta distribution
+    R3BDistribution<1>& GetBetaDistribution() { return fBeta; }
+    void SetBetaDistribution(const R3BDistribution<1>& betaDistribution) { fBeta = betaDistribution; }
+
+    // This method will calculate the beam beta using the given nucleon mass or approximate it from the number of
+    // nucleons
+    void SetEnergyDistribution(R3BDistribution<1> energyDistribution_AMeV, int nNucleons = 1, double mass_MeV = 0.);
+
+  private:
+    R3BDistribution<3> fVertex_cm;   //!
+    R3BDistribution<2> fSpread_mRad; //!
+    R3BDistribution<1> fBeta;        //!
+};
+
+#endif
--- a/r3bgen/R3BIonGenerator.cxx
+++ b/r3bgen/R3BIonGenerator.cxx
@@ -15,261 +15,117 @@
 // -----            Based on FairIonGenerator source file              -----
 // -----            Created 18/03/11  by M. Labiche                    -----
 // -------------------------------------------------------------------------
+
 #include "R3BIonGenerator.h"
 
+#include "FairIon.h"
 #include "FairLogger.h"
 #include "FairPrimaryGenerator.h"
-
-#include "FairIon.h"
 #include "FairRunSim.h"
 
 #include "TDatabasePDG.h"
-#include "TObjArray.h"
-#include "TParticlePDG.h"
-
 #include "TMath.h"
-#include "TRandom.h"
-
-#include "FairIon.h"
-#include "FairRunSim.h"
+#include "TObjArray.h"
 #include "TParticle.h"
-#include <iostream>
-using std::cout;
-using std::endl;
+#include "TParticlePDG.h"
+#include "TRandom.h"
+#include "TString.h"
 
-// -----   Initialsisation of static variables   --------------------------
-Int_t R3BIonGenerator::fgNIon = 0;
-// ------------------------------------------------------------------------
-
-// -----   Default constructor   ------------------------------------------
-R3BIonGenerator::R3BIonGenerator()
+R3BIonGenerator::R3BIonGenerator(UInt_t seed)
     : fMult(0)
-    , fPx(0.)
-    , fPy(0.)
-    , fPz(0.)
-    , fR(0.)
-    , fz(0.)
-    , fOffset(0.)
-    , fSigmaP(0.)
-    , fAngle(0.)
-    , fVx(0.)
-    , fVy(0.)
-    , fVz(0.)
-    , fIon(NULL)
-    , fQ(0)
-    , fBeamSpotIsSet(kFALSE)
+    , fIon(nullptr)
+    , fRNG(seed)
 {
-    //  cout << "-W- R3BIonGenerator: "
-    //      << " Please do not use the default constructor! " << endl;
 }
-// ------------------------------------------------------------------------
 
-R3BIonGenerator::R3BIonGenerator(const Char_t* ionName, Int_t mult, Double_t px, Double_t py, Double_t pz)
-    : fMult(0)
-    , fPx(0.)
-    , fPy(0.)
-    , fPz(0.)
-    , fR(0.)
-    , fz(0.)
-    , fOffset(0.)
-    , fSigmaP(0.)
-    , fAngle(0.)
-    , fVx(0.)
-    , fVy(0.)
-    , fVz(0.)
-    , fIon(NULL)
-    , fQ(0)
-    , fBeamSpotIsSet(kFALSE)
+R3BIonGenerator::R3BIonGenerator(const Char_t* ionName, Int_t mult, Double_t momentum_AGeV_per_c, UInt_t seed)
+    : fMult(mult)
+    , fIon(nullptr)
+    , fRNG(seed)
 {
+    fIon = (FairIon*)FairRunSim::Instance()->GetUserDefIons()->FindObject(ionName);
 
-    FairRunSim* fRun = FairRunSim::Instance();
-    TObjArray* UserIons = fRun->GetUserDefIons();
-    TObjArray* UserParticles = fRun->GetUserDefParticles();
-    FairParticle* part = 0;
-    fIon = (FairIon*)UserIons->FindObject(ionName);
-    if (fIon)
-    {
-        fgNIon++;
-        fMult = mult;
-        fPx = Double_t(fIon->GetA()) * px;
-        fPy = Double_t(fIon->GetA()) * py;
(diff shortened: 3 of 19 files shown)


r3bgen/R3BParticleGenerator.cxx
```diff
@@ -0,0 +1,137 @@
...
+};
+
+void R3BParticleGenerator::SetParticle(Int_t pdgCode)
+{
+    fPDGCode = pdgCode;
+    auto particle = TDatabasePDG::Instance()->GetParticle(pdgCode);
+    if (!particle)
+        LOG(fatal) << "R3BParticleGenerator: Unknown PDG Code!";
+
+    if (abs(pdgCode) > 1000000000)
```
janmayer, line 102: I would recommend not getting the masses for Ions from TDatabasePDG, but from Geant4, [like so](https://github.com/R3BRootGroup/R3BRoot/blob/8affc0f5150f0b8927f5c11a66cf29ceba0fed42/r3bgen/R3BPhaseSpaceGenerator.cxx#L52)
Also compare with the `AddHeavyIon` method, its been a while but `run->AddNewIon` might be necessary for the VMC layers to work with heavy ions.

Also, the SetIon/Particle interfaces differ between the different Generators - that might be a target to be unified.

r3bgen/R3BParticleGenerator.h
```diff
@@ -0,0 +1,103 @@
...
+#include "R3BDistribution.h"
+#include "R3BDistribution1D.h"
+#include "TRandom3.h"
+
+#include <iostream>
+
+class FairPrimaryGenerator;
+
+// C++17 will make the following lines much better, but for now we have to stick with this...
+// maybe not very good to read, but its at least constexpr
```
janmayer, line 27: I think this is already implemented either in ROOT or Geant4.

r3bgen/R3BParticleGenerator.h
```diff
@@ -0,0 +1,103 @@
...
+    R3BParticleGenerator();
+
+    R3BParticleGenerator(Int_t pdgCode, UInt_t seed = 0);
+    R3BParticleGenerator(const Char_t* ionName, unsigned int seed = 0);
+
+    void SetProton();
+    void SetNeutron();
+    void SetIon(const Char_t* ionName);
+    void SetParticle(Int_t pdgCode);
+    void SetParticle(Int_t charge, UInt_t mass)
```
janmayer, line 55: Instead of charge / mass rather Z, A to avoid confusion with mass in GeV/c².

Note that there is a huge overlap to the other generators, e.g. [`R3BPhaseSpaceGenerator`](https://github.com/R3BRootGroup/R3BRoot/blob/8affc0f5150f0b8927f5c11a66cf29ceba0fed42/r3bgen/R3BPhaseSpaceGenerator.cxx#L50) That might be something to extract and de-duplicate.

r3bgen/R3BParticleGenerator.h
```diff
@@ -0,0 +1,103 @@
...
+    {
+        fVertex_cm = vertexDistribution_x_y_z_cm;
+    }
+    R3BDistribution<3>& GetVertexDistribution() { return fVertex_cm; }
+
+    void SetEnergyDistribution(const R3BDistribution<1>& energyDistribution_AMeV)
+    {
+        fEnergy_AMeV = energyDistribution_AMeV;
+    }
+    R3BDistribution<1>& GetEnergyDistribution() { return fEnergy_AMeV; }
```
janmayer, line 75: Implicitly, the input is handled as AMeV for Ions and as MeV for particles. This does match the common usage and is the right behavior. Maybe add something to make it more explicit? Or a comment? idk.

janmayer (commented):
We're accumulating quite a number of different generators. There are some elements, especially with selecting the particles/ions that could be streamlined or de-duplicated into helper "mixins".

I'm not sure what to to with the R3BIonGenerator - it has slightly different functionality (multiple ions, not sure if anyone uses that). I would also like not having to drag it around forever for backwards compatibility.

r3bgen/R3BParticleGenerator.h
vadimr3b: Yeah I chose actually not change R3BIonGenerator, becasue it produces several ions, each from a different position, and not all from the same like I would have assumed.

r3bgen/R3BParticleGenerator.cxx
vadimr3b: > I would recommend not getting the masses for Ions from TDatabasePDG, but from Geant4, [like so](https://github.com/R3BRootGroup/R3BRoot/blob/8affc0f5150f0b8927f5c11a66cf29ceba0fed42/r3bgen/R3BPhaseSpaceGenerator.cxx#L52)

Does is not somewhat work against the abstraction?

> Also compare with the AddHeavyIon method, its been a while but run->AddNewIon might be necessary for the VMC layers to work with heavy ions.

It seems like its only a list of userdefined ions?

> Also, the SetIon/Particle interfaces differ between the different Generators - that might be a target to be unified.

Yes I agree. I would actually prefer to keep the number of generators small, but because of the (at least for me) special way R3BIonGenerator works, we have to decide if we want to break backwards compatibility or add a new very simple generator (this one basically).

janmayer: > Does is not somewhat work against the abstraction?

Yes, it does. But the abstraction is, uhm, "not great" anyhow.  It has been a while, but I had problems with the masses from ROOT - give it a try and compare the values. Note that in principle, Geant4 has all the masses already and also uses those (I think), so specifying them again (different source of truth) will only lead to tears - similar reason why the masses must not be specified in the ascii-input files. 

Personally, I would cut as much of the abstraction layer as possible and just use Geant4.


> It seems like its only a list of user defined ions?

Yes. Why that lists exists and especially, why it needs (might be needed?) to be populated manually, I don't know.

vadimr3b: > Yes, it does. But the abstraction is, uhm, "not great" anyhow. It has been a while, but I had problems with the masses from ROOT - give it a try and compare the values. Note that in principle, Geant4 has all the masses already and also uses those (I think), so specifying them again (different source of truth) will only lead to tears - similar reason why the masses must not be specified in the ascii-input files.

Well FairRoot will use TDatabasePDG to get the masses if you pass the corresponding flag.
I feel like, that this might be a problem, but can (or rather should) not be solved here.

> Yes. Why that lists exists and especially, why it needs (might be needed?) to be populated manually, I don't know.

I added the lines. The next push will have them included.

vadimr3b:
I now moved particle selection and beam properties selection to their own classes, making the code much more reusable and the individual generators simpler to understand, since only a few lines of code remained there.

I switched from beam energy to beam beta, since this is actually what we want.

janmayer:
> I switched from beam energy to beam beta, since this is actually what we want.
Hm, I liked setting the Energy in AMeV, as its much easier to talk and reason about. Now I would have to calculate the beta from outside the class if I want to simulate at eg, 200, 600, and 1000 AMeV

vadimr3b:
>  Hm, I liked setting the Energy in AMeV, as its much easier to talk and reason about. Now I would have to calculate the beta from outside the class if I want to simulate at eg, 200, 600, and 1000 AMeV

Yes, same for me.
I am thinking about a simple way to set it up.
Like this the problem of calculating the right beta is gone.
Otherwise we need the energy in AMeV, the number of nucleons and the mass of the particle.
I could write a methoed, which will take a beam energy distribution and put it into a beta distribution approximation, assuming nucleon mass etc.
What do you think about that?

janmayer:
The masses etc. are required, but I think that is fine, we need those anyway. What was the problem with calculating beta on the fly if needed? [Like so?](https://github.com/R3BRootGroup/R3BRoot/blob/9da7eb913007c6ee08d5844cc865f6d97273066e/r3bgen/R3BPhaseSpaceGenerator.cxx#L80)

janmayer:
I have to think about that some more. First, I think its not right to derive from `R3BParticleSelector` and `R3BBeamPropertiesSelector`. That reads as: "The Phase Space Generator is a Fair Generator and a Particle Selector and a Beam Properties Selector", which is wrong. I feel the right grammar would be "The Phase Space Generator is a Fair Generator and **has** a Particle Selector and a Beam Properties Selector".

That could translate into a usage like 
```
R3BPhaseSpaceGenerator gen;
gen.particles.add(<pdg code>);
gen.beam.setEnergy(<distribution>);
```
or something like that.

vadimr3b:
I added a Methoed to R3BBeamPropertiesSelector.
 
> I have to think about that some more. First, I think its not right to derive from `R3BParticleSelector` and `R3BBeamPropertiesSelector`. That reads as: "The Phase Space Generator is a Fair Generator and a Particle Selector and a Beam Properties Selector", which is wrong. I feel the right grammar would be "The Phase Space Generator is a Fair Generator and **has** a Particle Selector and a Beam Properties Selector".
> 
> That could translate into a usage like
> 
> ```
> R3BPhaseSpaceGenerator gen;
> gen.particles.add(<pdg code>);
> gen.beam.setEnergy(<distribution>);
> ```
> 
> or something like that.

Well I would see this as a common interface to setup the beam properties.
Since you cannot define interfaces in C++ this feels more natural for me.
The way you suggest would not have any benefits in our case:
For the BeamPropertiesSelector:
We would have only one BeamPropertiesSelector and I cannot see any other usefull scenario?
This means beeing able exchaning it would be useless.
To avoid 

-  Use getters/setters would ruine the whole purpose of the class.
-  Exposing it would make possible reactions after appy

vadimr3b:
> The masses etc. are required, but I think that is fine, we need those anyway. What was the problem with calculating beta on the fly if needed? [Like so?](https://github.com/R3BRootGroup/R3BRoot/blob/9da7eb913007c6ee08d5844cc865f6d97273066e/r3bgen/R3BPhaseSpaceGenerator.cxx#L80)

Well it works, but it is just an approximation, which has obviously some error for protons and neutrons.
I added a method to `R3BBeamPropertiesSelector` with the option to actually get rid of the error in case of protons etc.

>I have to think about that some more. First, I think its not right to derive from R3BParticleSelector and R3BBeamPropertiesSelector. That reads as: "The Phase Space Generator is a Fair Generator and a Particle Selector and a Beam Properties Selector", which is wrong. I feel the right grammar would be "The Phase Space Generator is a Fair Generator and has a Particle Selector and a Beam Properties Selector".
>
>That could translate into a usage like
>
>R3BPhaseSpaceGenerator gen;
>gen.particles.add(<pdg code>);
>gen.beam.setEnergy(<distribution>);
>
>or something like that.

Well I would see this as a common interface to setup the beam properties of a generator.
Since you cannot define interfaces in C++ this feels more natural for me.

For the `R3BBeamPropertiesSelector`:
We would have only one BeamPropertiesSelector and I cannot see any other useful scenario, this means it would basically just be a namespace and add a few more letters in a few lines.
Do not see the benefit in such small classes, but hey separating it just to separate is okay for me.

For the `R3BParticleSelector`:
I think here the idea is to map a broad and common external interface to a narrow internal one.
I could imagine, that this could be used with some restrictions for the selected particles.
E.g. only ions and in case you add a gamma, it throws an exception or whatever.
Doing this immediately would tell the user in the stack trace where the problematic particle was added instead pointing to some init function.
I think that this might end up with data duplication and lots of unnecessary lines of code.
But again, this could be done.

janmayer (changes requested):
Have you tried that the mass determination from ROOT works?

```
root [0] TDatabasePDG::Instance()->GetParticle(2112)->Mass();
(double) 0.93956500
root [1] TDatabasePDG::Instance()->GetParticle(1000501320)->Mass();
<crash>
```

I'm also using the Geant4 Database in [R3BAsciiGenerator.cxx](https://github.com/R3BRootGroup/R3BRoot/blob/72202c7838812e992592712c59054655fe935885/r3bgen/R3BAsciiGenerator.cxx#L190). Should also have higher precision.

vadimr3b:
> Have you tried that the mass determination from ROOT works?
> 
> ```
> root [0] TDatabasePDG::Instance()->GetParticle(2112)->Mass();
> (double) 0.93956500
> root [1] TDatabasePDG::Instance()->GetParticle(1000501320)->Mass();
> <crash>
> ```
This is a dealbreaker.
I changed the code a bit and moved the things around.
The mass determination ins `R3BParticleSelector` uses now `G4NistManager`.
The classes should be intuitive and easy to use.

r3bgen/R3BParticleSelector.h
```diff
@@ -0,0 +1,46 @@
...
+    void AddParticle(const char* particleName,
+                        int z,
+                        int a,
+                        int s,
+                        double mass_GeV,
+                        int charge,
+                        bool stable,
+                        double decaytime_ns);
+    void AddParticle(int pdgCode);
+    void AddParticle(int z, uint a);
```
janmayer, line 40:
```suggestion
    void AddParticle(int z, int a);
```

r3bgen/R3BParticleSelector.cxx
```diff
@@ -0,0 +1,103 @@
...
+
+    // for other particles TDatabasePDG should be fine
+    const auto particle = TDatabasePDG::Instance()->GetParticle(pdgCode);
+    if (!particle)
+        LOG(fatal) << "R3BParticleSelector: Unknown PDG Code!";
+
+    addParticle(pdgCode, particle->Mass());
+}
+
+void R3BParticleSelector::AddParticle(int z, uint a)
```
janmayer, line 95:
```suggestion
void R3BParticleSelector::AddParticle(int z, int a)
```

r3bgen/R3BParticleSelector.cxx
```diff
@@ -0,0 +1,103 @@
...
+#include "TParticlePDG.h"
+
+#include "G4NistManager.hh"
+
+constexpr auto PDG_Proton = 2212;
+constexpr auto PDG_Neutron = 2112;
+
+// C++17 will make the following lines much better, but for now we have to stick with this...
+// maybe not very good to read, but its at least constexpr
+constexpr int ChargeMass2PDGCode(int z, uint a)
```
janmayer, line 31:
```suggestion
constexpr int ChargeMass2PDGCode(int z, int a)
```

r3bgen/R3BBeamProperties.cxx
```diff
@@ -0,0 +1,40 @@
...
+R3BBeamProperties::R3BBeamProperties()
+    : fVertex_cm(R3BDistribution3D::Delta(0, 0, 0))
+    , fSpread_mRad(R3BDistribution2D::Delta(0, 0))
+    , fBeta(R3BDistribution1D::Delta(0.5))
+{
+}
+
+void R3BBeamProperties::SetEnergyDistribution(R3BDistribution<1> energyDistribution_AMeV, int nNucleons, double mass)
+{
+    const auto nucleonMass = (mass == 0. ? nNucleons * AMU_MeV : mass);
```
janmayer, line 31: You could also use the Particle Selector (or just the mass lookup) here, but this should be good enough.

janmayer (changes requested):

r3bgen/R3BBeamProperties.h
```diff
@@ -0,0 +1,59 @@
...
+    {
+        fSpread_mRad =
+            R3BDistribution<2>({ spreadDistribution_theta_mRad, R3BDistribution1D::Flat(0, 2 * TMath::Pi() * 1e3) });
+    }
+
+    // beta distribution
+    R3BDistribution<1>& GetBetaDistribution() { return fBeta; }
+    void SetBetaDistribution(const R3BDistribution<1>& betaDistribution) { fBeta = betaDistribution; }
+
+    // This method will calculate the beam beta by assuming nucleon mass.
```
janmayer, line 48:
```suggestion
```

janmayer (commented):
Looks good to me, now just update the test [`testR3BPhaseSpaceGeneratorIntegration`](https://github.com/R3BRootGroup/R3BRoot/blob/dev/r3bgen/test/testR3BPhaseSpaceGeneratorIntegration.C) with the new method names.

r3bgen/R3BBeamProperties.h
vadimr3b: I will wait two more days in case @kresan or someone else has also other suggestions, then I will commit those changes and the changes in the macros aswell.

janmayer: Note that the test is in the main repository <sub><sup>(as it should be)</sup></sub>, not in the macro repository - to get the travis tasks passing.

vadimr3b:
I tried to adapt the `R3BIonGenerator` to the new changes.
I basically mostly only removed code and fixed a few problems.

Is everybody actually aware, that this generator does not do what you think it does?
If you set the multiplicity to ten, it will produce the then particles all at the same position with same momentum.

r3bgen/R3BIonGenerator.cxx
```diff
@@ -15,261 +15,114 @@
...
-    fPz = Double_t(a) * pz;
-    // fVx   = vx;
-    // fVy   = vy;
-    // fVz   = vz;
-    char buffer[20];
-    sprintf(buffer, "FairIon%d", fgNIon);
-    fIon = new FairIon(buffer, z, a, q);
-    FairRunSim* run = FairRunSim::Instance();
+    ++fgNIon;
+    fIon = new FairIon(TString::Format("FairIon%d", fgNIon), z, a, q);
```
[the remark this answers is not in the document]
vadimr3b: I think this is not connected to the error.
The problem was that FairRoot internally got the mass from `TDatabasePDG` if the mass was not passed.
I fixed this.

r3bgen/R3BIonGenerator.cxx
```diff
@@ -15,261 +15,117 @@
 // -----            Based on FairIonGenerator source file              -----
```
janmayer, line 15: That comment block could be removed

r3bgen/R3BIonGenerator.h
```diff
@@ -22,9 +22,13 @@
 #ifndef R3BIONGENERATOR_H
```
janmayer, line 22: The comments above could be removed

janmayer (approved):
Code is fine. I'm not really satisfied with the class names `IonGenerator` and `ParticleGenerator`, as they do different things that can't be inferred from the names, but oh well.

PR #368, opened 2021-01-19
PSPX: bug fixes in PSP classes and R3BPspxOnline

sonjastorck:


kresan:
you need to fix the clang-format issues and force-push again

PR #379, opened 2021-02-10
Fixes #376 Check if gtest is found (which it should be!)

### LINKED ISSUE #376: Make dependency on gtest optional
@janmayer Do you use gtest? I have just noticed that it is not in nov20 release of FairSoft? Can we make this dependency optional?

janmayer:


kresan:
I have asked about googletest in FairSoft in a dedicated channel. Will try to get it fixed

PR #382, opened 2021-02-17
Added Boost::regex lib to califa CMakeList file

jose-luis-rs:


kresan:
Maybe also with this PR remove the loading of regex library in the CALIFA code

hapol:
If the only use of the regex library is for the search of a pattern in the name of the crystals within R3BCalifaGeometry, we could revert to the previous code version (check fixed position in the char*) which is simpler, faster and should not produce any error as the char buffer is fixed by the geometry macro.

PR #415, opened 2021-05-11
Add missing #include for newer ROOT / C++

ChristianTackeGSI:
I was trying to compile with newer versions of ROOT and C++17 (namely: apr21 / Fairroot 18.6).

It looks like some more `#include`s are needed.
Those don't seem to hurt on jun19 / Fairroot 18.2.
So let's do this!


ChristianTackeGSI:
Thanks!

PR #419, opened 2021-05-27
Implemented the reading of runids as function of timestamps

jose-luis-rs:


r3bbase/R3BFileSource.cxx
```diff
@@ -357,6 +375,45 @@ Bool_t R3BFileSource::Init()
         fInputFile.seekg(0, std::ios::beg);
     }
 
+    if (fInputFile.is_open())
+    {
+        LOG(INFO) << "R3BFileSource::Init() Reading RunId file";
+        std::string buffer;
+        char* p;
```
bl0x, line 382: array_size should be explicitly set to 0 here, before starting the loop.

r3bbase/R3BFileSource.h
```diff
@@ -205,11 +197,21 @@ class R3BFileSource : public FairSource
...
+
+    /** input file with runids */
     std::ifstream fInputFile;
     TString fInputFileName;
-    Int_t ReadIntFromString(const std::string& wholestr, const std::string& pattern);
-    R3BEventHeader* fEventHeader;
-    Int_t fEntryMax;
+
+    UInt_t fexpid;
+    Int_t array_size;
```
bl0x, line 208: This should be called fArraySize, should it not?

r3bbase/R3BFileSource.cxx
```diff
@@ -357,6 +375,45 @@ Bool_t R3BFileSource::Init()
...
+            if (buffer.find("FILE") == 0)
+                continue;
+
+            getline(fInputFile, buffer);
+            LOG(INFO) << "read from file: \"" << buffer << "\"";
+
+            std::string buf1 = buffer.c_str();
+            buf1.erase(buf1.find(' '), buf1.length());
+            buffer.erase(0, buffer.find(' ') + 1);
+            fexpid = strtol(buf1.c_str(), &p, 16);
```
bl0x, line 403: You should check, that after a call to `strtol` the value returned in `p` is different from the value passed in as `buf1.c_str()`. If that is not the case (i.e. if `p == buf1.c_str()` is true), then no value was converted.

r3bbase/R3BFileSource.cxx
```diff
@@ -357,6 +375,45 @@ Bool_t R3BFileSource::Init()
...
     }
 
+    if (fInputFile.is_open())
+    {
+        LOG(INFO) << "R3BFileSource::Init() Reading RunId file";
+        std::string buffer;
+        char* p;
+        do
+        {
+            getline(fInputFile, buffer);
```
bl0x, line 385: The return value of `getline` should be checked: If it returns -1, then something went wrong, and `errno` indicates the error.

r3bbase/R3BFileSource.cxx
bl0x: This also applies to the other invocations of `strtol`

r3bbase/R3BFileSource.cxx
```diff
@@ -357,6 +375,45 @@ Bool_t R3BFileSource::Init()
...
+            LOG(INFO) << "read from file: \"" << buffer << "\"";
+            if (buffer.find("fRunID") == 0)
+            {
+                buffer.erase(0, buffer.find('=') + 1);
+
+                frunid[array_size] = strtol(buffer.c_str(), &p, 16);
```
bl0x, line 391: It is not checked, if array_size is still within the allowed limits. Being an `Int_t`, it may as well have a negative value, or a value larger than the number of elements allocated for `frunid`.

r3bbase/R3BFileSource.cxx
```diff
@@ -357,6 +375,45 @@ Bool_t R3BFileSource::Init()
...
+            getline(fInputFile, buffer);
+            LOG(INFO) << "read from file: \"" << buffer << "\"";
+
+            std::string buf1 = buffer.c_str();
+            buf1.erase(buf1.find(' '), buf1.length());
+            buffer.erase(0, buffer.find(' ') + 1);
+            fexpid = strtol(buf1.c_str(), &p, 16);
+            ftimestamp[array_size] = strtol(buffer.c_str(), &p, 16);
+            LOG(DEBUG) << fexpid << " / " << ftimestamp[array_size];
+            array_size++;
```
bl0x, line 406: This is not safe against overflow of `array_size`, which would result in a negative value (and is undefined behaviour in any case).

r3bbase/R3BFileSource.cxx
```diff
@@ -370,38 +427,49 @@ void R3BFileSource::SetInTree(TTree* tempTree)
     Init();
 }
 
+// ----  Method getrunid -----------------------------------------------------
+Int_t R3BFileSource::getrunid(uint64_t st)
+{
+    for (Int_t i = 0; i < array_size - 1; i++)
```
bl0x, line 433: When using `std::vector`, this loop can be replaced with a call to the `find_if` algorithm.

Also, one would assume that the timestamps in the file are sorted from smaller to larger values, correct? Also, the events are processed in increasing order, correct?
If that is true, then one can store the index into the array/vector (called `i` here) that points to the current timestamp. And then in the next call to getrunid start start from the index of the last timestamp without having to go through all previous ones.

r3bbase/R3BFileSource.cxx
```diff
@@ -370,38 +427,49 @@ void R3BFileSource::SetInTree(TTree* tempTree)
...
 {
     /** TODO
      ** We should use here the timestamp from the header to look for the right runId
      ** and set up the parameters for the analysis of the root files.
      ** std::cout << fEvtHeader->GetTimeStamp() << std::endl;
-     ** Below an example with the event number, just for tests.
      **/
 
-    if (i > fEntryMax && fEntryMax != -1 && fInputFile.is_open())
+    if (nextts == 0 && fEvtHeader->GetTimeStamp() > nextts)
```
bl0x, line 464: Can't one contract these two `if` and `if else` blocks into a single one with the condition `nextts >= 0` ?
Also, what should happen, when the condition is not true? Is there an `else` block that may be helpful?

jose-luis-rs:
Thank you Bastii, at the moment this is for test. I have observed that in the lmd files there are many events that have the same timestamp, which is not expected for me. Do you know why?

Cheers

PR #420, opened 2021-05-31
Added runid to the R3BEventHeaderCal2Hit class for its propagation

jose-luis-rs:
nothing

Added whiterabbit propagator for data analysis

Update of R3BFileSource class

Cosmetic changes

Cosmetic changes

Cosmetic change

Update readme

### DIFF
--- a/r3bbase/R3BFileSource.cxx
+++ b/r3bbase/R3BFileSource.cxx
@@ -45,6 +45,8 @@ using std::set;
 #include <iostream>
 using namespace std;
 
+TMCThreadLocal R3BFileSource* R3BFileSource::fSourceInstance = 0;
+
 R3BFileSource::R3BFileSource(TFile* f, const char* Title, UInt_t)
     : FairSource()
     , fInputTitle(Title)
@@ -81,15 +83,16 @@ R3BFileSource::R3BFileSource(TFile* f, const char* Title, UInt_t)
     , fTimeProb(0)
     , fCheckFileLayout(kTRUE)
     , fInputFile()
-    , fexpid(0)
-    , array_size(0)
+    , fExpid(0)
+    , prevts(0)
     , nextts(0)
 {
-    for (Int_t j = 0; j < 99999999; j++)
+    if (fSourceInstance)
     {
-        frunid[j] = 0;
-        ftimestamp[j] = 0;
+        Fatal("R3BFileSource", "Singleton instance already exists.");
+        return;
     }
+    fSourceInstance = this;
 
     if (fRootFile->IsZombie())
     {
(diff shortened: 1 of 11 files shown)


r3bbase/R3BFileSource.cxx
```diff
@@ -400,10 +422,19 @@ Bool_t R3BFileSource::Init()
...
+
+            if (buf1.c_str() == p2 || buffer.c_str() == p3)
+            {
+                LOG(ERROR) << "R3BFileSource::Init() Bad structure for RunId file, \n"
+                           << "it must be (hex format): \n"
+                           << "fRunID = a \n"
+                           << "expid timestamp \n";
+                return kFALSE;
+            }
+            fTimestamp.push_back(ts);
```
bl0x, line 436: I think it makes sense here, since you are using two vectors now, that you check after pushing the timestamp, that the sizes of fTimestamp and fRunid are equal. Otherwise the two vectors are running out of sync (perhaps due to some bug introduced in the future). Just a sanity check.

I've seen the sync check further down, but it is better to (also) check here in every iteration of the loop, because two errors (one in timestamp and one in runid) may cancel each other out.

r3bbase/R3BFileSource.cxx
```diff
@@ -427,26 +458,38 @@ void R3BFileSource::SetInTree(TTree* tempTree)
...
 }
 
-// ----  Method getrunid -----------------------------------------------------
-Int_t R3BFileSource::getrunid(uint64_t st)
+// ----  Method GetRunid -----------------------------------------------------
+Int_t R3BFileSource::GetRunid(uint64_t st)
 {
-    for (Int_t i = 0; i < array_size - 1; i++)
-        if (st >= ftimestamp[i] && st < ftimestamp[i + 1])
+    UInt_t fArray = fTimestamp.size();
```
bl0x, line 464: Better use 'auto' type for fArray. Name suggestion: fArray -> fArraySize or fNumTimestamps or fNumEntries.

r3bbase/R3BFileSource.cxx
```diff
@@ -461,15 +504,16 @@ Int_t R3BFileSource::ReadEvent(UInt_t i)
...
+           100. * i / (double)(fNoOfEntries),
+           fRunId);
+    fflush(stdout);
+
+    if (nextts >= 0 && prevts >= 0 && (fEvtHeader->GetTimeStamp() > nextts || fEvtHeader->GetTimeStamp() < prevts))
     {
-        fRunId = getrunid(fEvtHeader->GetTimeStamp());
-        fEvtHeader->SetRunId(fRunId);
+        fRunId = GetRunid(fEvtHeader->GetTimeStamp());
     }
```
bl0x, line 517: What should fRunId be set to, if the condition is false?

jose-luis-rs: It will take the last value for the fRunId, or if the timestamp is not defined inside the runid file it will be set to 1.

r3bbase/R3BFileSource.cxx
jose-luis-rs: Yes, you are right! Thanks.

PR #431, opened 2021-07-14
added proof of concept 1 wrts per califa hit unpacker

klenze:


r3bdata/califaData/R3BCalifaMappedData.h
```diff
@@ -16,46 +16,32 @@
...
-  protected:
-    UShort_t fCrystalId; // crystal unique identifier
-    Int_t fEnergy;       // total energy in the crystal
-    Int_t fNf;           // total fast amplitude in the crystal
-    Int_t fNs;           // total slow amplitude in the crystal
-    uint64_t fTime;      // time-stamp (common to all the hits in the event)
-    UChar_t fError;      // bit coded error flag
-    Int_t fTot;          // time-over-treshold
-
+#define MEMBER_PLUS_GETTER(type, name)                                  \
```
hapol, line 23: Could you please write the explicit data members and get functions in the interface?
Code is going to be written once but read by many people many times. There is no need for this additional complexity which obscure the code for many people.

jose-luis-rs: Yes, I agree with Hector.
I will adapt the code in the dev branch and check that everything works for the s455 experiment.

klenze: I was trying to avoid the duplication between getter return types and variable types, but this is obviously better solved by N3638, e.g. auto GetFoo {return fFoo;}. Preprocessor stuff is always a crutch. 

But per [C++ core guidelines C.131](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-get), we should probably refactor this class into a struct anyhow.

PR #442, opened 2021-08-16
Compatibility with FairRoot 18.4+

janmayer:
- `R3BINCLRootGenerator.cxx`: Added missing include
- `FairMCTracks` has beend removed from FairRoot. Here, remove unused references and roll code into `R3BMCTracks`
- `R3BFileSource.cxx`: Add a define to deal with ActivateObjectAnyImpl, added in FairRoot 18.4
- `R3BFileSource.h`: `SpecifyRunId` is not implemented, which leads to crashes. As it is required by the base class, provide a dummy implementation marked as fixme.

Note: Why does `R3BFileSource` even exist in the first place? It is only used in `R3BEventHeaderCal2Hit` for some Run Id shenanigans. Please solve this in a way that does not create duplication with `FairFileSource`, i.e. by deriving from `FairFileSource` instead of `FairSource`. The naming of `R3BEventHeaderCal2Hit` does not match its function. My guess is that some functionality of FairRoot is misued, misunderstood, or ignored and `R3BEventHeaderCal2Hit` and `R3BFileSource` should be deleted.

### DIFF
--- a/evtvis/R3BCalifaEventDisplay.cxx
+++ b/evtvis/R3BCalifaEventDisplay.cxx
@@ -19,7 +19,6 @@
 
 #include "FairEventManager.h"
 #include "FairLogger.h"
-#include "FairMCTracks.h"
 #include "FairRootManager.h"
 #include "FairRunAna.h"
 #include "FairRuntimeDb.h"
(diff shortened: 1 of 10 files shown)


jose-luis-rs:
R3BFileSource was created to manage our detector parameters as a function of the timestamps, which cannot be done from FairFileSource. I created this class to make some tests. In my last PR, I have also updated this class as you proposed.

I have also changed the name of R3BEventHeaderCal2Hit to R3BEventHeaderPropagator. This class is needed to propagate the runid info from root files with raw data to next data levels (Cal, Hit, etc). Some examples about how to use these classes are in the SOFIA repository for the s455 experiment: https://github.com/R3BRootGroup/sofia/tree/dev/macros/s455Up2p/unpack

PR #481, opened 2021-12-20
updated los, sci2 classes; added new AnalysisIncomingID

MartinaFeijoo:
califa julich struct and reader also


kresan:
please rebase

PR #482, opened 2021-12-23
Compatibility with new event header name R3BEventHeader -> EventHeader.

igasparic:
ext_h101 file for 25 planes
GainMatching vtarget.A -> vtarget


kresan:
@igasparic your change in the tamex mapping file breaks some NeuLAND tests. This happens because of mismatch between the new mapping file and old sample data file. Please create a PR with the corresponding example data file in the macros repository.

kresan:
duplicate of #485. closing

PR #488, opened 2022-01-18
Implement a basic skeleton for RPC data structures and reconstructors

hapol:
Includes the r3bdata/rpcData structures (Mapped, Cal, Hit) according to the naming
conventions and a set of calibrators to move from a level to the next. It is also
included a prototype for the R3BRpc class for simulation and the point structure.
In this case, the connection should be checked with the geo files (volume names
should match the suggestions in this code for sensitive volumes).


kresan:
Please resolve the conflicts and rebase.

PR #515, opened 2022-03-10
Add Delta DT to R3BMusliOnline

audreychatillon:
clang-format


audreychatillon:
Small debugging from the on-line analysis.
Fix a segmentation fault while resetting the MUSLI_HISTO

jose-luis-rs:
The online does not work, please, check all the index again. The problem seems in the init function but I don't get to find the problem.

jose-luis-rs:
Please, accept this PR.
Thanks in advance

PR #521, opened 2022-03-14
Updated analysis and online for FOOTs

morze00:


jose-luis-rs:
WR info was deleted in the online, but we can add it later. Please, accept this PR for the current test.
Thanks in advance

jose-luis-rs (approved):

Issue #535, opened 2022-03-29
Multiple conditions for skipping events based on Tpat values


ryotani:
I noticed there are two independent conditions for selecting events with the Tpat and they are interfering.
The first definition can be found in R3BUcesbSource.cxx, here the condition aims to eliminate meaningless events with tpat=0.:
[https://github.com/R3BRootGroup/R3BRoot/blob/a558707338e1c61dc3ee83298c1f2a4b6b1fbad4/r3bsource/base/R3BUcesbSource.cxx#L292-L302](url)

>     if (fSkip)
>     {
>         if (fEventHeader->GetTpat() > 0)
>         {
>             FairRunOnline::Instance()->MarkFill(kTRUE);
>         }
>         else
>         {
>             FairRunOnline::Instance()->MarkFill(kFALSE);
>         }
>     }

While in R3BTrloiiTpatReader.cxx, another condition can be found as:
[https://github.com/R3BRootGroup/R3BRoot/blob/a558707338e1c61dc3ee83298c1f2a4b6b1fbad4/r3bsource/trloii/R3BTrloiiTpatReader.cxx#L102-L112](url)

>     if (fTpat >= 0 && fEventHeader && fTrigger <= 0)
>     {
>         if ((fEventHeader->GetTpat() & fTpat) == fTpat)
>         {
>             FairRunOnline::Instance()->MarkFill(kTRUE);
>         }
>         else
>         {
>             FairRunOnline::Instance()->MarkFill(kFALSE);
>         }
>     }

The latter definition aims to select desired trigger condition by setting fTpat value, but when we set fSkep=true in R3BUcesbSource, it will overwrite the condition of FairRunOnline::Instance()->MarkFill(). It is not a big issue but it would be great to remove such confusion. How can we resolve this? Should we write a condition to check the fTpat value of R3BTrlolliReader in R3BUcesbSource as well as we check the fSkip value?

PR #538, opened 2022-04-01
Event selection with Tpat conditions with R3BTrloiiTpatReader will be…

ryotani:
… prioritised to R3BUcesbSource as the former can include the condition of the latter.
This should answer the issue #535 .

A bug in R3BMusicMapped2Cal.cxx to get slopes for the calibration parameter fixed.
Small typos in mwpc classes are corrected.


jose-luis-rs:
Thanks Ryo.

PR #570, opened 2022-05-11
MusliCal2Hit : load the HitPar and TGeoPar in ContFact

audreychatillon:
change MusliCalData and MusliHitData formats

MusliHitPar class implemented

Add Cal2Hit step with Z and theta calculation

debug


audreychatillon:
I fetch and rebase again in case that was the problem

PR #602, opened 2022-05-21
S522 califa proton online histograms with different cuts. Quick & Dirty.

klenze:
Online histograms related to the maximum energy cluster hits for either side. 

For now, i have implemented a few cuts based on TPats. 

A very UNIX-spirited commit ("Do one thing ~~and do it well~~", "Worse is Better")

### DIFF
--- a/califa/CMakeLists.txt
+++ b/califa/CMakeLists.txt
@@ -67,6 +67,7 @@ set(SRCS
 ./online/R3BCalifaOnlineSpectra.cxx
 ./online/R3BCalifaJulichOnlineSpectra.cxx
 ./online/R3BCalifaDemoOnlineSpectra.cxx
+./online/R3BCalifaProtonHists.cxx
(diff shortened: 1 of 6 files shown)


gabrigarjim:
Just a simple question,  is this class a bit specific to this experiment? Most probably the tpats that are used here for sorting are  not going to be the same again....

klenze:
@gabrigarjim yes, but it could easily be adapted to other experiments, so it's scope should be broader than e.g.  R3BCalifaJulichOnlineSpectra. 

It is also meant as a proof of concept that you do not necessarily have to pack all of your hists in a single class, with hundreds of lines between declaration, allocation and filling of histograms. 

One way to generalize it would be to add the desired tpat cuts after instantiating the class. 
Or you could add a virtual bool Filter() method in which the user can make cuts on anything by overloading the method using an anonymous class. 

In general, it would be nice to have access to the tpat names (and downscaling settings), but I don't think this is in the lmd.

PR #620, opened 2022-06-03
Updated the way to treat multiplicity at S2 in R3BSci2Tcal2Hit. Assum…

ryotani:
…ing the pair of hits are giving the right time if the multiplicities for both left and right are the same, then adding all to Sci2HitData. Then by R3BIncomingBeta will find a good hit comparing the beta gates.


jose-luis-rs:
Thanks Ryo

jose-luis-rs:
Please, review your PR with data from the current experiment because it ends with "*** Break *** segmentation violation"

PR #647, opened 2022-08-04
Update hit s2

nmozumdar:
Update hit selection in S2

### DIFF
--- a/sci2/calib/R3BSci2Tcal2Hit.cxx
+++ b/sci2/calib/R3BSci2Tcal2Hit.cxx
@@ -44,6 +44,7 @@ R3BSci2Tcal2Hit::R3BSci2Tcal2Hit(const char* name, Int_t iVerbose)
     , fsci2OffsetXT(0.)
     , fClockFreq(1. / VFTX_CLOCK_MHZ * 1000.)
     , fOnline(kFALSE)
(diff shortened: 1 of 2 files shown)


ryotani:
Although I have no time to follow carefully the logic, as long as I read all changes once, I think the update looks fine. 
One comment:
You defined COINC_WINDOW for selecting a "good hit" within one plastic detector to have a reasonable time difference between left and right PMTs. Can you make it a parameter at least accessible from a mutator function as SetCoincWindow(Float_t tdiff), or even set from the parameter class?

nmozumdar:
Yes I can definitely implement this, I guess using a function is easier and makes it more accessible. Thanks a lot for the comments.

ryotani:
If you make an update, then you just push the new commit to your branch, then it should automatically update the pull request. It's not urgent but mentioning this just in case if you struggle how to deal with github.

nmozumdar:
Thanks a lot for the advice, I am a complete novice to git hub so any help is appreciated. I have made the changes

Btw I needed to talk to you about s2 calibration and hit selection  using s2-los, I am running some checks now and i will send you a mail shortly after

ryotani:
Yes, feel free to send me an email. Or we can discuss it in the issues section of this github repository.

PR #667, opened 2022-10-06
PSPX1 detector added for incoming ID, condition to ensure incoming To…

ivanalihtar:
…Fs are in same VFTX coarse cycle added.


jose-luis-rs:
I am doing it, including other details as well...

nmozumdar:
Ok, thanks a lot.

ivanalihtar:
Dear Nikhil and Jose Luis,

sorry for the mistake and thanks for the correction.

Best, Ivana

PR #672, opened 2022-10-10
Added MDF wrapper class R3BMDFWrapper for tracking through GLAD

morze00:


tracking/TrackingLinkDef.h
```diff
@@ -27,6 +27,7 @@
 #pragma link C++ class R3BTrackingDetector+;
 #pragma link C++ class R3BTrackingParticle+;
 #pragma link C++ class R3BTrackingSetup+;
+#pragma link C++ class R3BMDFWrapper++;
```
jose-luis-rs, line 30: Why ++ ?

tracking/R3BMDFWrapper.cxx
```diff
@@ -0,0 +1,215 @@
...
+
+R3BMDFWrapper::~R3BMDFWrapper()
+{
+}
+
+//This method will read a special txt file with MDF parameters, don't change
+void R3BMDFWrapper::InitMDF(const char* mdf_file)
+{
+    int i;
+    std::ifstream file(mdf_file);
```
hapol, line 40: Is there any reason to use an input file when they can be loaded as parameters using the standard parameters management in R3BRoot/FAIRRoot? The use of ifstream does not allow control by the RunTimeDB and the files are not included in the parameters repositories... If there is no good reason for that, I would prefer to create a parameter container for this information.

tracking/R3BMDFWrapper.cxx
```diff
@@ -0,0 +1,215 @@
...
+        std::cout << "\nUnable to open MDF parameter file!!";
+    return;
+
+}
+
+void R3BMDFWrapper::PrintMDF()
+{
+    std::cout << "\n\n======== Initialized MDF from the R3BMDFWrapper:";
+
+    std::cout << std::setprecision(9) << "\nmdf_NVariables = "  << mdf_NVariables
```
hapol, line 83: It does not use the FAIRRoot/R3BRoot LOG standard which would be preferible for comments (it includes function and line where the call is made).

tracking/R3BMDFWrapper.cxx
```diff
@@ -0,0 +1,215 @@
...
+        std::cout << mdf_Power[i] << ", ";
+
+    std::cout << "\n\n";
+}
+
+//This method will read a special txt file with PCA parameter, don't change
+void R3BMDFWrapper::InitPCA(const char* pca_file)
+{
+    int i;
+    std::ifstream file(pca_file);
```
hapol, line 120: Again, it would be preferible to have all in parameter containers

hapol (changes requested):

tracking/TrackingLinkDef.h
morze00: sorry, just a typo... will be fixed

tracking/R3BMDFWrapper.cxx
morze00: thanks, I will fix all messages to use the usual LOG standard

tracking/R3BMDFWrapper.cxx
morze00: Would it be possible to accept the current functionality temporarily? I will prepare the appropriate parameter container later when the rest of the code (tracking task) is developed. Due to some technical reasons (the way how this input file is currently generated, tested, readout etc.) I would prefer to preserve this part of the code.  For sure there will be several other changes in the future and I will try to make everything using standard parameter containers.  At the moment the MDFWrapper class  is necessary for the further code development and several people are waiting for this to be added to the dev branch...

tracking/R3BMDFWrapper.cxx
hapol: It's ok for me... do you want to change first the LOG vs cout issue and correct the typo? Then we will accept the PR...

tracking/R3BMDFWrapper.cxx
hapol: ok, no problem

tracking/R3BMDFWrapper.cxx
morze00: Yes, working on it right now :) I will make soon a new pull request where the typo and LOG issues are fixed

PR #673, opened 2022-10-15
Fixing a potential segmentation error from R3BUcesbSource

YanzhaoW:
If the event header in FairRunOnline is not set to R3BEventHeader before its initialisation, FairRun would create an event header in type `FairEventHeader` and register it in `FairRootManager`. Such automatic creation of event header happens after the initialisation of source in `FairRunOnline::Init()`. Therefore, we can checking if R3BEventHeader is already set or not in `R3BUcesbSource::Init()`. If not, we set it to the R3BEventHeader(not owned). Note that registration of event header is not needed here since `FairRunOnline` will register it later.

### DIFF
--- a/r3bsource/base/R3BUcesbSource.cxx
+++ b/r3bsource/base/R3BUcesbSource.cxx
@@ -61,6 +61,22 @@ R3BUcesbSource::~R3BUcesbSource()
 
 Bool_t R3BUcesbSource::Init()
 {
+    // Register of R3BEventHeader in the output root file
+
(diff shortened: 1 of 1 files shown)


YanzhaoW:
> Please, squash the commits by using the command:
> 
> > git rebase -i mainrepo/dev
> > Details in:
> > https://github.com/AnarManafov/GitWorkflow/blob/master/GitWorkflow.markdown

Sorry. Now it's done.

jose-luis-rs:
Thanks

PR #679, opened 2022-10-25
Remake of Califa Hit Finder

gabrigarjim:
Remake of Califa Hit finder class. Some new features :

- Addition of Get Crystal List method. Retrieves a list of crystals that belong to the cluster :
`std::vector<Int_t> crystalList` 
`crystalList = hit->GetCrystalList()`
**Edit : GetNbOfCrystalHits() is still working** 

The first element of the list is the one that was taken as a cluster center candidate (the one with more energy inside the cluster). That one is also the crystal that gives angular information.

- Cluster Thresholds. There are three different clusters : gammas , protons and saturated (with a gamma saturated crystal as cluster candidate). Each one has its own energy threshold. Typical Usage : 
  
   `R3BCalifaCrystalCal2Hit* CalifaCal2Hit = new R3BCalifaCrystalCal2Hit();`
     `CalifaCal2Hit->SetRandomization(kTRUE);`
     `CalifaCal2Hit->SetRandomizationFile("angular_histograms.root");`
     `CalifaCal2Hit->SetClusterWindow(0.25);`
     `CalifaCal2Hit->SetCrystalThreshold(200); `
     `CalifaCal2Hit->SetGammaClusterThreshold(800);`
     `CalifaCal2Hit->SetProtonClusterThreshold(50000);`
     `CalifaCal2Hit->SetSaturationEnergy(30000);`
     `CalifaCal2Hit->IsSimulation(kFALSE);`

- Is Simulation flag.
- Cluster Types. Gamma, proton and saturated clusters are retrieved by :
`Int_t clusterType = hit -> GetClusterType()` (0 for protons, 1 for gammas and 2 for saturated )

**Edit : Added suggestions by @YanzhaoW  and @tcsousa99**

### DIFF
--- a/califa/calibration/R3BCalifaCrystalCal2Hit.cxx
+++ b/califa/calibration/R3BCalifaCrystalCal2Hit.cxx
@@ -30,33 +30,100 @@
 
 #include "R3BCalifaCrystalCalData.h"
 #include "R3BCalifaGeometry.h"
+#include "R3BCalifaMappingPar.h"
 #include <list>
 #include <vector>
 
-#include "ROOT_template_hacks.h"
-using roothacks::TCAHelper;
-using roothacks::TypedCollection;
+using namespace std;
+
+struct califa_candidate {
+
+  Int_t motherId;
+  vector<Int_t> crystalList;
+  Double_t energy;
+  Double_t ns;
+  Double_t nf;
+  Double_t theta;
+  Double_t phi;
+  ULong64_t time;
+
+};
+
+bool compareByEnergy(R3BCalifaCrystalCalData *a , R3BCalifaCrystalCalData *b)
+{
+    return a->GetEnergy() < b->GetEnergy();
+}
+
+
+bool isInside(vector<Int_t> &vec, Int_t cryId)
+{
+    bool result = false;
+    if( find(vec.begin(), vec.end(),cryId) != vec.end() )
+    {
+        result = true;
+    }
+    return result;
+}
+
+
+void RemoveUsedCrystals(vector<Int_t> &used , vector<R3BCalifaCrystalCalData*> &all , vector<R3BCalifaCrystalCalData*> &proton, vector<R3BCalifaCrystalCalData*> &gamma, vector<R3BCalifaCrystalCalData*> &saturated){
+
+   for(int p = 0 ; p < used.size() ; p++){
+    for(int s = 0 ; s < all.size() ; s++)
+
+      if(all.at(s)->GetCrystalId() == used.at(p))
+       all.erase(all.begin() + s);
+
+     }
+
+   for(int p = 0 ; p < used.size() ; p++){
+    for(int s = 0 ; s < gamma.size() ; s++)
+
+      if(gamma.at(s)->GetCrystalId() == used.at(p))
+       gamma.erase(gamma.begin() + s);
+
+     }
+
+   for(int p = 0 ; p < used.size() ; p++){
+     for(int s = 0 ; s < proton.size() ; s++)
+
+      if(proton.at(s)->GetCrystalId() == used.at(p))
+       proton.erase(proton.begin() + s);
+
+     }
+
+
+   for(int p = 0 ; p < used.size() ; p++){
+     for(int s = 0 ; s < saturated.size() ; s++)
+
+      if(saturated.at(s)->GetCrystalId() == used.at(p))
+       saturated.erase(saturated.begin() + s);
+
+     }
+
+
+}
+
 
 R3BCalifaCrystalCal2Hit::R3BCalifaCrystalCal2Hit()
+
     : FairTask("R3B CALIFA CrystalCal to Hit Finder")
     , fCrystalCalData(NULL)
     , fCalifaHitData(NULL)
-    , fGeometryVersion(2021) // BARREL+iPhos
-    , fThreshold(0.)         // no threshold
-    , fDRThreshold(15000)    // in keV, for real data (15000 = 15MeV)
-    , fDeltaPolar(0.25)
-    , fDeltaAzimuthal(0.25)
-    , fDeltaAngleClust(0)
-    , fClusterAlgorithmSelector(RECT)
-    , fParCluster1(0)
+    , fGeometryVersion(2021)
+    , fCrystalThreshold(100.) // 100 keV
+    , fProtonThreshold(15000) // Crystal saturation @ ~15 MeV
+    , fGammaClusterThreshold(1000) // 1 MeV
+    , fProtonClusterThreshold(50000) // 50000 MeV
+    , fMap_Par(NULL)
+    , fRoundWindow(0.25)             // 14 degrees
+    , fSimulation(kFALSE)
     , fCalifaGeo(NULL)
-    , fNbCrystalsGammaRange(2432) // during 2019 it was 5000
     , fOnline(kFALSE)
     , fRand(0)
     , fRandFile("")
 {
-    SetSquareWindowAlg(fDeltaPolar, fDeltaAzimuthal);
-    fCalifatoTargetPos.SetXYZ(0., 0., 0.);
+
 }
 
 R3BCalifaCrystalCal2Hit::~R3BCalifaCrystalCal2Hit()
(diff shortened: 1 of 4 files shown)


califa/calibration/R3BCalifaCrystalCal2Hit.cxx
```diff
@@ -98,224 +156,553 @@ InitStatus R3BCalifaCrystalCal2Hit::Init()
...
+  usedCrystals.push_back(motherId);
+
+
+
+  // First we add all proton hits
+  for(Int_t j = 0 ; j < allCrystalVec.size() ; j++ ){
+
+    Int_t thisCryId = allCrystalVec.at(j)->GetCrystalId();
+
+    if(thisCryId==motherId)
```
klenze, line 288: not required, would otherwise be caught by !isInside(usedCrystals,thisCryId))

klenze (commented):
I was the one who wrote the previous version of the algorithm, so I am not exactly unbiased.  Some comments.

My version took about 100 lines, with the actual clustering algorithm making up 50 lines of that (If one took out the comments, debug logging and your randomization which IMHO really does not belong into the middle of our most fundamental data processing step, I end up with just eighteen LOC). The actual clustering code used just a single data structure, std::list<R3BCalifaCrystalCalData*>. The relevant invariant (either a CCCalData is in a cluster or in the list) were rather obvious (to me). 

Your version of the of clustering takes 315 nonempty lines (if blocks are started in a new line). You have five different std::vector<CCCalData*> and one std::vector<int> which you have to keep consistent.

Further remarks:

* I guess that we decided that 2432 is the threshold for all time and nobody cares about the 2019 beam times anyhow?

* The complexity of my version is obviously O(n^2). The complexity of your version is a bit harder to calculate. I think the bottleneck is the following: You have O(n) calls to RemoveUsedCrystals (assuming you have n/const clusters). In that function, you loop over used (O(n), because you never clear it) and then over all (also O(n)). The erase call is expensive (O(n)) but will only be called O(n) times in total, so we can disregard it. From my calculation, the comparison in line 74: if(all.at(s)->GetCrystalId() == used.at(p)) will be called O(n^3) times in total.

* Is fMap_Par->GetInUse etc safe to call with out of range crystal IDs?

* While I agree that removing the myriads of strange neighborhood definitions is probably a good thing, I do not agree that the cone is the fairest method. The distance from the target to a crystal is much shorter for theta=90deg than for the GLAD-wards part of the barrel. A cone wide enough to encompass the next neighbors at theta=90deg will probably encompass many more crystals for the forward barrel. My favorite algorithm would be to just pick crystals within a given radius of the crystal which created the cluster.

* One of the key points for having the simulation and the experiment analysis in the same code base is that one can use the same analysis on both. If you treat the simulation in a completely different branch (which will invariably diverge further with time), that defeats that purpose. Also, your code gets twice as long. 

* I still would have left the match function alive, even if it is just a one-liner for now. Later, we might want to create separate classes for various neighborhood definitions. 

* From my understanding, the goal of your PR is to merge low-gain hits preferably with other low gain hits, instead of picking the range dependent on the energy the individual crystal saw. Why do you assume this to be an improvement over the present behavior?

* Assuming fSaturation > fGammaClusterThreshold, RemoveUsedCrystals should have emptied saturatedCandidatesVec because the original saturatedCandidatesVec was a subset of gammaCandidatesVec?

* en>fSaturation is not a good way to check for overflows. Last I recalled, I marked them with en=nan. It might be a good idea to change that to en=infty.

* What is our endgame for overflow clusters anyhow? My approach would have been:
** For dual range crystals for which we have data from both ranges, select a range based on the low-gain range being above or below a threshold.
** Set overflows to infinite.
** For the entries with infinite energy, figure out if there is a ToT measurement. If yes, use that as the energy instead.
** Having created a unique best energy estimate for each crystal, do the clustering. 

* As far as I can remember, the point about using TClonesArrays was that people were arguing that malloc/free would be so slow. Putting a data member which clearly uses dynamically allocated memory in them kind of defeats that argument.

* Speaking of memory, I assume that you have tested your code for memory leaks? It seems entirely possible that the destructor for the R3BCalifaHitData's member std::vector<Int_t> fCrystalList never gets called because you use Clear() in the Reset() method.

Given my usual level of agreement with the R3BRoot maintainers, I expect that your code will be merged without further discussion as soon as one of them scrolls through my complaints. :-)

On the plus side, it is nice to see people using STL containers instead of TObjectArray or whatever.

klenze:
Ah, seems like format checks are a bit less strict nowadays:
Checking against parent commit HEAD^
fatal: ambiguous argument 'HEAD^': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
✓ passed clang-format checks clang-format-8

YanzhaoW:
Maybe a general remark about the coding style. I see the `Exec()` member function has hundreds of lines of code. It's great if this works. But it could be better if some parts could be implemented inside some other member functions. And then call those member functions in `Exec()`. It would look more modular, readable and maintainable. :D

gabrigarjim:
Hi, I will try to reply to all of your remarks (sorry for the delay )

> My version took about 100 lines, with the actual clustering algorithm making up 50 lines of that (If one took out the comments, debug logging and your randomization which IMHO really does not belong into the middle of our most fundamental data processing step, I end up with just eighteen LOC). The actual clustering code used just a single data structure, std::list<R3BCalifaCrystalCalData*>. The relevant invariant (either a CCCalData is in a cluster or in the list) was rather obvious (to me).

As I told you previously I think that the randomization is an important feature to some people doing analysis and that step well placed into a hit data structure (in my opinion). Anyways, if someone does not like it is just as simple as _SetRandomization(false)_ . Respect to code lenght, I agree that maybe some parts of it can be condensed into auxiliary functions. 

> I guess that we decided that 2432 is the threshold for all time and nobody cares about the 2019 beam times anyhow?

While I agree that this part of the Califa software needs an improvement, I do not agree with your approach to that (If I remember well, you wanted to introduce a new data member in mapped & cal levels. Even if you use a bool then it will add a total of 0.4% of extra data. That seems to be innocent, but if CALIFA spits ~40 tb of data (like in my experiment) that implies like ~150 gb of unnecessary data + proccesing time). My approach would be to load the total number of crystals of the califa version (2019, 2022, whatever) in the mappingPar  and use it as a threshold in the algorithms.  

> The complexity of my version is obviously O(n^2). The complexity of your version is a bit harder to calculate. I think the bottleneck is the following: You have O(n) calls to RemoveUsedCrystals (assuming you have n/const clusters). In that function, you loop over used (O(n), because you never clear it) and then over all (also O(n)). The erase call is expensive (O(n)) but will only be called O(n) times in total, so we can disregard it. From my calculation, the comparison in line 74: if(all.at(s)->GetCrystalId() == used.at(p)) will be called O(n^3) times in total.

I have not calculated the complexity in detail, but I made some test over real data comparing your previous algorithm and this one. For an standard (p,2p) analysis both methods were quite similar in performance  but if I wanted to perform some gamma analysis by forming gamma clusters and low thresholds mine was faster. If the CALIFA daq is working properly, we expect to see at max like 20-30 crystals per event (as we saw in Anna's experiment this year) so I don't see that the difference in performance is noticeable in a normal case. 

> Is fMap_Par->GetInUse etc safe to call with out of range crystal IDs?

You mean to call it with a non existing crystal ID or a non installed one? In any case I can write an small check to check this kind of things, thats a good idea.

> While I agree that removing the myriads of strange neighborhood definitions is probably a good thing, I do not agree that the cone is the fairest method. The distance from the target to a crystal is much shorter for theta=90deg than for the GLAD-wards part of the barrel. A cone wide enough to encompass the next neighbors at theta=90deg will probably encompass many more crystals for the forward barrel. My favorite algorithm would be to just pick crystals within a given radius of the crystal which created the cluster.

This I agree. I just used the cone one as a first implementation. The idea of this pull request (and the califa wg  meeting that took place several weeks ago) was to encourage the users to ask for their desired functionalities. So if you think that the radius one is a good addition then I will include it, of course.  On the other hand I think that the cone is quite good, as the forward part is the one that is receiving particles with highest energies and therefore a wider window would be desired in that region. 

> One of the key points for having the simulation and the experiment analysis in the same code base is that one can use the same analysis on both. If you treat the simulation in a completely different branch (which will invariably diverge further with time), that defeats that purpose. Also, your code gets twice as long.

Saturated crystals and double ranged ones made me think about different approaches. I agree that both analysis should be as similar as possible, but in the simulation we do not have a saturation in a crystal and maybe this has to be treated differently. 

> I still would have left the match function alive, even if it is just a one-liner for now. Later, we might want to create separate classes for various neighborhood definitions.

This can be done, yes. 

> From my understanding, the goal of your PR is to merge low-gain hits preferably with other low gain hits, instead of picking the range dependent on the energy the individual crystal saw. Why do you assume this to be an improvement over the present behavior?

In the present algorithm there is only a crystal threshold for everything. That makes you to choice between processing everything and last forever (for example in a dirty channel) or exclude low energy hits and miss gamma information. The goal of this pull request is then to separate between the two main analysis in CALIFA (gamma and proton) , include new features (list of crystals of a given cluster) and add as many new features as users want, so we can be prepared for the times to come. 

Sorry to say, but the e = nan approach is not working, as we can se some nice saturation peaks around 30 MeV by selecting gamma ranges and plotting the energy that we get. Anyways, the goal of this was to be able to select proton hits even if we are hitting only-gamma-read channels. This is useful for cross sections calculations, ToT reconstruction, etc... 


> As far as I can remember, the point about using TClonesArrays was that people were arguing that malloc/free would be so slow. Putting a data member which clearly uses dynamically allocated memory in them kind of defeats that argument.

Sorry, but at the times this was discussed I was probably starting high school :) (jokes aside, I usually prefer to work with std vectors)

> Speaking of memory, I assume that you have tested your code for memory leaks? It seems entirely possible that the destructor for the R3BCalifaHitData's member std::vector<Int_t> fCrystalList never gets called because you use Clear() in the Reset() method.

There are no memory leaks as far as I know. That vector may not be removed, but as it is cleared after every event then the only problem would be at the end of the unpacking, but it would be only the size of the empty vector, right?

> Given my usual level of agreement with the R3BRoot maintainers, I expect that your code will be merged without further discussion as soon as one of them scrolls through my complaints. :-)

That's not quite true. You know that if the comments are constructive ones then things are discussed and solved. From my side the plan now is to wait for your complaints/ideas/suggestions, close the PR, add everything important that I miss or correct mistakes and upload this again. 

Side note : Is a bit annoying to have to PR modified macros to make your PR in the main repo work , given that you are changing the class that is used in the test, so I completely agree with your proposal (https://github.com/R3BRootGroup/R3BRoot/issues/688#). @kresan , is there any reason that I am missing why macros is not inside the main repo?

gabrigarjim:
Closing PR to update with suggestions.

PR #682, opened 2022-10-28
Update of Neuland validated data class

YanzhaoW:
The validation class in `R3BRoot/neuland/shared` folder is used to wrap the data with a validated tag. User can only get the value of the data if it's validated. This is very useful if the data value needs heavy calculations. And the recalculation of the data is only required if the data is invalidated, which could improves the efficiency of algorithms.

Following changes of the `Validated` class can further improves the efficiency, especially when data is very large and copying becomes very expensive.

### DIFF
--- a/neuland/shared/Validated.h
+++ b/neuland/shared/Validated.h
@@ -22,6 +22,7 @@
 // D Modified by Jan Mayer
 
 #include <assert.h>
+#include <utility>
 
 //=============================================================================
 template <typename T>
@@ -32,7 +33,8 @@ class Validated
     // invalid constructor
 
     Validated(const T& data);
-    // valid constructor
+    Validated(T&& data);
+    // valid constructors
 
     bool valid() const;
     // is this valid or not?
@@ -42,10 +44,14 @@ class Validated
     // cached value.
 
     void set(const T& data);
+    void set(T&& data);
     // set the data, makes the instance valid.
 
     T get() const;
     // explicit conversion
+
+    const T& getRef();
+    // getter with const ref. Use this if you only want to check the value.
 
     operator T();
     operator T() const;
@@ -84,6 +90,17 @@ Validated<T>::Validated(const T& data)
 
 //=============================================================================
 template <typename T>
+Validated<T>::Validated(T&& data)
+    //
+    // D Valid constructor
+    //
+    : m_valid(true)
+    , m_data(std::move(data))
+{
+}
+
+//=============================================================================
+template <typename T>
 bool Validated<T>::valid() const
 //
 // D is this valid or not?
@@ -116,7 +133,26 @@ void Validated<T>::set(const T& data)
 
 //=============================================================================
 template <typename T>
+void Validated<T>::set(T&& data)
+//
+// D set the data.
+//
+{
+    m_data = std::move(data);
+    m_valid = true;
+}
+
+//=============================================================================
+template <typename T>
 T Validated<T>::get() const
+{
+    assert(valid() && "Cannot return invalid data.");
+    return m_data;
+}
+
+//=============================================================================
+template <typename T>
+const T& Validated<T>::getRef()
 {
     assert(valid() && "Cannot return invalid data.");
     return m_data;


klenze:
I am by no means an expert with rvalue references and std::move, but to me, it looks like (your version)
```
void set(T data)
{
    m_data = std::move(data);
}
```
and (previous version)
```
set(const T& data)
{
   m_data = data;
}
```
will both require one deep copy of the object. In your version, this happens when the object is passed on the call stack in preparation of the function call, in the previous version, it happens when the assignment in the function is executed. 

Furthermore, I suppose that ``m_data = std::move(data);`` implies a shallow copy (copying of sizeof(T) bytes) while pass by reference takes copying sizeof(T*) bytes, which is typically smaller. 

I would assume that one could get a single shallow copy if one added 
```
void set(T&& data)
{
    m_data=data; // I guess this will call operator=(T&&) even without std::move
}
```
and calling it using ``set(std::move(myobj))``. Of course, this will reset myobj.

std::move sees very little little use outside of neuland (and I would give even odds that the few uses in r3bdata and r3bgen were also written by @janmayer). The rest of R3BRoot (and, to be honest, ROOT) uses raw pointers. Of course, T* has a whole bunch of meanings:
* Transfer of ownership
* Pass by reference
* A 1d C-array, passed by reference
** with the length implied by zero termination (C-strings, argv in execv(3))
** with the length provided elsewhere (memcpy(3))
* Pass by reference, read only implied (it's not like we are const correct)

In some places, pointers will probably remain a long time (though we could technically replace T* with smart pointers). In other places, we might slowly move to references. I personally feel that C arrays should be limited to dealing with external functions which require them (looking at you, ROOT). 

``\begin{semi-off-topic} %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%``
For a fun overview of R3BRoot's use of post-1983 C++ features try ``grep --exclude-dir=.git 'operator=' -r . `` in the source directory.
Outside of neuland and some small parts of califa, you will find two "implementations" of operator=:
* ``private: const R3BSci2HitPar& operator=(const R3BSci2HitPar&);``, with no implementation anywhere
  * which means that people using ``#define public private`` will get linker errors (which are, like everything, even more [fun](https://dwarffortresswiki.org/index.php/DF2014:Losing) in root macros, which might be acceptable.
  * I guess that before C++11 intruduced ``operatorFoo(...)=delete;``, this was an okayish way to mark an operator which you did not want to provide?
** I still would have preferred an implementation which asserted 0 or threw an exception. 
* ``public: R3BMfiPoint& operator=(const R3BMfiPoint&) { return *this; } ``
  * Having personally been bitten by this, I think this is neither fair nor sporting. 
  * No, the defense "I was only copying code without understanding what it did" won't fly.
  * Technically, It should be ``R3BMfiPoint& operator=(const R3BMfiPoint&) const``, as the assignment does not change the object being assigned to.

Surprisingly, we use lambda expressions in a couple of places though. 
``\end{semi-off-topic}``

I have no idea why that califa CI task failed btw.

Looking forward to more C++ discussion in the future,
   Philipp

janmayer:
> std::move sees very little little use outside of neuland (and I would give even odds that the few uses in r3bdata and r3bgen were also written by @janmayer). 

👋 

I have this https://youtu.be/6lurOCdaj0Y in my notes from back then, not sure if that helps.

YanzhaoW:
@klenze Thanks for your comment.
## About std::move
By use of 
```cpp
void set(T data)
{
    m_data = std::move(data);
}
```
it's like killing two birds with one stone: store the data by copying and store the data by moving.  The latter choice is impossible with previous version. So the principle is:
-  if you want to store a large date, use `set(std::move(data))`. 
- If you want to store a small data, use `set(data)`.

Yes, you can also explicitly overload the setter function with r reference. But why not just write a single setter function which can do as exactly the same job as two setter functions?

There is another problem with your suggested version:
```cpp
void set(T&& data)
{
    m_data=data; // I guess this will call operator=(T&&) even without std::move
}
```
It will not call operator=(T&&) because even though parameter of the setter is r-value, the named variable `data` is still l-value. If you want to prevent expensive copy, you still need `m_data=std::move(data)`.

## About shallow copy:
In my perspective, shallow copy happens when an object, which has a pointer type class member, is copied to another object. During this shallow copy, only the value of the member pointer (aka the address) is copied rather than the object, to which the member pointer is pointing. With this, all copy assignments perform shallow copies. 

The thing with `m_data=std::move(data)` is that std::move can only perform "move" if  `data` is movable and `data` contains members/attributes which can also be movable. If not, `m_data=std::move(data)` is basically the same as `m_data = data`. Therefore, yes, std::move could also have shallow copy if the object is not movable. 

But if one carefully designs his code, such as making the member pointer encapsulated inside smart pointers, moving the object would trickle down to the movement of smart pointers. Thus, not need to worry about the shallow copy.

I also don't understand why the test failed to pass with califa. All Neuland tests work fine.

YanzhaoW:
@jose-luis-rs 
Hi, 

I changed back to the unrevised version and the test still fails. Is there something wrong with the original CI testing?

jose-luis-rs:
Hello

The problem is due to some modifications in the git-repo macros, could you check it using the commit #55 of that repository?
https://github.com/R3BRootGroup/macros/tree/58036d9505409edf7e700a30d23373e31a73a811

Thanks in advance

klenze:
Okay, I did some experiments.
[test.cxx.txt](https://github.com/R3BRootGroup/R3BRoot/files/9903245/test.cxx.txt)

Suppose ```sizeof(test)``` was N and the total memory requirements of the instance was M. 
Then (for sufficiently large N, M) the cost of test's constructor/operator= taking an rvalue reference should be N, while the cost of the constructor/operator= taking an lvalue reference should be M.

Different types have very different distributions of N and M. A ``std::vector<int>`` (on my box) has N=24 and M (roughly) proportional to its entries.

By contrast, a ``std::array<int, 1000>`` has N=M=4000. 

```
Calls made with JanSet(obj):
  Called test& test::operator=(const test&)

Calls made with yourSet(std::move(obj)):
  Called test::test(test&&)
  Called test& test::operator=(test&&)

Calls made with yourSet(obj):
  Called test::test(const test&)
  Called test& test::operator=(test&&)

Calls made with mySet(obj):
  Called test& test::operator=(const test&)

Calls made with mySet2(obj):
  Called test& test::operator=(test&&)
```

* Jan's version has a cost of M. 
* yourSet(std::move(obj)) has a cost of 2*N
* yourSet(obj) has a cost of N+M
* yourSet(std::move()) has a cost of 2*N
* mySet(std::move(obj)) has a cost of M, because my assumption in the comment was wrong (probably something about the zoo of reference types)
* mySet2(std::move(obj)) finally has a cost of N. 

(All of this assumes, of course, the object being passed has the proper assignment and constructor types.)

If N is small (e.g. your objects is mostly pointers to dynamically allocated memory) and M is big, then your version with std::move should indeed be faster than Jan's version. mySet2(std::move(obj)) will probably be the fastest in any case, at the cost of always resetting the argument passed. 

I think if you want to both be maximally efficient and support move/nonmove, you need to implement two versions of set, e.g. mine and Jan's. This is what the STL [generally does](https://en.cppreference.com/w/cpp/container/vector/push_back).  

--

Regarding the non-operation assignment, I think some version of ROOT probably required operator= to be defined, and this was the quick fix they came up with. Which will be while std::sort(std::vector<R3BMfiPoint>, ...) will not do anything, for example. 

I agree that it is preferable to catch stuff at compile time if you can. Otherwise, it is still much better to catch them at runtime than just silently have them do unexpected stuff (e.g. nothing) silently.

--

Also, having the macros in a different repository yet still require them to pass the CI tests is fun, isn't it? However, if one wanted to optimize for painfulness, I would suggest making macros a git submodule of R3BRoot.

YanzhaoW:
@klenze 

Some questions:

- Where (or how) did you get the values of cost?

- Better use std::vector<int> since std::array<int> is not movable.

klenze:
> Where (or how) did you get the values of cost?

Are you asking for my reasoning behind:
> Then (for sufficiently large N, M) the cost of test's constructor/operator= taking an rvalue reference should be N, while the cost of the constructor/operator= taking an lvalue reference should be M.

Or how I go from 
```
  Calls made with JanSet(obj):
  Called test& test::operator=(const test&)
```
to
> Jan's version has a cost of M.

Re std::array, I do not advocate it generally over std::vector (but think it is still better than C arrays). I was merely pointing out that depending on your underlying type T, the expected runtime performance of the different solutions could be vastly different.

YanzhaoW:
@klenze 
>Then (for sufficiently large N, M) the cost of test's constructor/operator= taking an rvalue reference should be N
What do you mean by "cost of test"? If you mean the additional memory for the copying, move constructor/assignment should have cost of 0.

`std::array` and `std::vector` are used in different scenario. One cannot simply replace another. But since `std::vector` can be moved, it's well suited to store large data.

klenze:
> What do you mean by "cost of test"? If you mean the additional memory for the copying, move constructor/assignment should have cost of 0.

Okay, I was a bit unclear. I meant "if test had any data members thus that sizeof(test)=N and the total memory requirements of a typical used test instance are M, then (for sufficiently large N, M) the time cost of test's constructor/operator= taking an rvalue reference should be N, while the cost of the constructor/operator= taking an lvalue reference should be M."

YanzhaoW:
> > What do you mean by "cost of test"? If you mean the additional memory for the copying, move constructor/assignment should have cost of 0.
> 
> Okay, I was a bit unclear. I meant "if test had any data members thus that sizeof(test)=N and the total memory requirements of a typical used test instance are M, then (for sufficiently large N, M) the time cost of test's constructor/operator= taking an rvalue reference should be N, while the cost of the constructor/operator= taking an lvalue reference should be M."

Hi, I am still confused with the "cost". You said N is the size of test in bytes. But later you said "the time cost should be N", which is in seconds. Is cost you are mentioning the cost of memory or the cost of time?

klenze:
Next version:
If test had any data members thus that sizeof(test)=N and the total memory requirements of a typical used test instance are M bytes, then the time cost of test's constructor/operator= taking an rvalue reference should be O(N), while the cost of the constructor/operator= taking an lvalue reference should be O(M).

YanzhaoW:
> Next version:
> If test had any data members thus that sizeof(test)=N and the total memory requirements of a typical used test instance are M bytes, then the time cost of test's constructor/operator= taking an rvalue reference should be in O(N), while the cost of the constructor/operator= taking an lvalue reference should be in O(M).

ok, now I understand. But are you  only talking about unmovable large data or both movable and unmovable large data?

jose-luis-rs:
Hi, did you finish this PR?

YanzhaoW:
> Hi, did you finish this PR?

Hi, yes, I did the final change to this PR. But somehow the error persists.

PR #693, opened 2022-11-08
helpful warning

klenze:


klenze:
I would claim that the failure to build is intentional, but from the output, it looks rather like it crashes and burns even before my error message is triggered.

jose-luis-rs:
The master branch was updated two months ago:
https://github.com/R3BRootGroup/R3BRoot/releases/tag/sep22
so this PR is a bit strange ...

klenze:
At the moment, to the best of my knowledge, nobody is recommending the master branch for anyone. I just tried compiling it, which fails with the TDatime for me. 

If we want to have the master branch as a kind of "rolling release" (and using tags for releases) which contains code which is better tested than dev, we at least should quickly backport bug fixes. I do not think we have the developer time to do that.

PR #699, opened 2022-11-16
Create an executable for Neuland simulation

YanzhaoW:
So this is the first step to replace ROOT Macro with C++ executable.

To use this executable:
```sh
cd build && cmake ..
make
source config.sh # this mush be done before running the executable
cd bin
./neuland
```
### Attention
 There is a problem to this if you run it in lxir136. Since we are now using C++17 with 
```sh
source /u/land/fake_cvmfs/sourceme.sh nov22
```
the dynamic libraries and the executable are linking to the new C++ standard library in the folder`/u/land/fake_cvmfs/9.13/gcc-10.2.0/lib64`. If you don't include this to your `${LD_LIBRARY_PATH}` env variable, it will use the old version of C++ library in `/usr/llb`, which could make program fail. (this will be fixed in the future.)

So in this case, simply include it with:
```sh
export LD_LIBRARY_PATH=/u/land/fake_cvmfs/9.13/gcc-10.2.0/lib64:${LD_LIBRARY_PATH}
```

### Looking for some help
I strongly don't like sourcing the shell script before running the executable:
```sh
source config.sh
```
What an executable should do is: open the terminal and execute the program and that's it! No need of caring about env variables and sourcing some files. 

I'm all ears if anyone has any solution to this.


klenze:
Personally, I would name the binary something like neuland_sim_test. Just because we do not have any other binaries at the moment one can still hope that at one day we might -- perhaps a neuland sim which takes arguments on how many events you want to simulate, or a neuland data unpacker. 

With regard to the LD lib path, I think you can try to convince the linker to link to the absolute path of libstdc++.so. (This may get hacky: may have to tell g++ not to automatically link to it and instead figure out its full path and feed that to the linker. Kids, don't do this when you want to create binaries which can be shared between machines.)

config.sh should imho set everything needed to run the binary. If it does not write all the contents of (ld) lib path during compilation in front of ld lib path, consider complaining to FairRoot. 

Getting rid of the requirement of sourcing config.sh to get binaries seems like a tricky thing to do. I am not sure how Geant4 searches for the reaction data, for example. Perhaps SIMPATH is baked in during compilation, or perhaps it is evaluated during runtime. Or  the file``R3BRoot/config/rootmanager.dat`` -- is it used at all? When is it evaluated? Looking at your main, I would say VMCWORKDIR definitely needs to be set.

Also, are you loading any configuration settings from *.C or *.yaml? In the past, I was unable to compile scripts which tried to load gConfig.C.

YanzhaoW:
@klenze 

Thanks for your comments

> Personally, I would name the binary something like neuland_sim_test. Just because we do not have any other binaries at the moment one can still hope that at one day we might -- perhaps a neuland sim which takes arguments on how many events you want to simulate, or a neuland data unpacker.

This is the very first step. In the future, I'm planning to use this executable to perform multiple tasks, including simulation, FairTasks and data analysis, depending on the options or configurations in YAML file.

>config.sh should imho set everything needed to run the binary. If it does not write all the contents of (ld) lib path during compilation in front of ld lib path, consider complaining to FairRoot.

I once complained to them about env variables set by FairRoot messed up with other important applications, such as ssh and git. They said they are going to stop setting up LD_LIBRARY_PATH. See [issue](https://github.com/FairRootGroup/FairSoft/issues/471). So, in this case, telling them to include more paths to the env variable seems not very applicable. And in the FairRoot github issues, you can also see people complaining about FairRoot setting up two many env variables.

>Also, are you loading any configuration settings from *.C or *.yaml? In the past, I was unable to compile scripts which tried to load gConfig.C.

Yes, if you run the executable, you can see the output 
```
User path for Configuration (g4Config.C) is used
```

>With regard to the LD lib path, I think you can try to convince the linker to link to the absolute path of libstdc++.so.

How can I do that with CMake?

YanzhaoW:
Another comment about "configuration settings from *.C or *.yaml":

Actually I had exactly the same problem before. The reason is even though I don't use ROOTCINT but the binary/executable, FairRoot still uses ROOTCINT to interpret that gConfig.C config file. And if your env variables are not set correctly to let ROOTCINT use the correct libraries, it will fails to interpret the gConfig.C file.

klenze:
If bin/neuland is to be general purpose, I have no issue with the naming. What will you be using to parse command line arguments? boost::program_options? Or is there a nice parser library which can handle both hierarchical yaml and (flat?) argv somewhere?

I am looking at the linker options in the gcc man page. I think the least painful way (rather than going ``-nodefaultlibs`` and manually adding all the default libs again) would be to use ``-static-libstdc++``. You probably have to set that in CMAKE_EXE_LINKER_FLAGS. Run with make VERBOSE=1 to see if the flags are actually used.

YanzhaoW:
> If bin/neuland is to be general purpose, I have no issue with the naming. What will you be using to parse command line arguments? boost::program_options? Or is there a nice parser library which can handle both hierarchical yaml and (flat?) argv somewhere?
> 
> I am looking at the linker options in the gcc man page. I think the least painful way (rather than going `-nodefaultlibs` and manually adding all the default libs again) would be to use `-static-libstdc++`. You probably have to set that in CMAKE_EXE_LINKER_FLAGS. Run with make VERBOSE=1 to see if the flags are actually used.

Exactly, I use boost::program_options, which has been already included in FairRoot. This is good enough. For yaml config file, I use yaml-cpp. I can't see any necessity to parse them together.

Actually, I changed my mind and will change the naming of the executable for the next PR.

Will `-static-libstdc++` link the standard libs to the binary file in static way, which makes the binary executable much bigger?

klenze:
> Will `-static-libstdc++` link the standard libs to the binary file in static way, which makes the binary executable much bigger?

Well, the libstdc++ is 17MiB for gcc 9.4. That is about the size of the neuland libraries. So while it is certainly not great, it would also not in my top 500 things to optimize about R3BRoot :-)

But yes, if you were to create 50 ELF files, static linking libstdc++ might not be the way to go.

jose-luis-rs:
@YanzhaoW did you finish with this PR?

YanzhaoW:
@jose-luis-rs yes, please merge this.

Issue #705, opened 2022-11-28
PSA: Califa WRTS, Overflows


klenze:
Dear all, 
it has come to my attention that the old unpacking code (``FEBEX3_CALIFA_BASE()`` in the spec file) does not only not encode the individual hits WRTS, but also does not include the overflow flags. This will mean that overflows will not get invalidated in R3BCalifaMapped2CrystalCal (to ieee 754 NAN, possibly INFINITY in the future) but erroneously get propagated as valid data. 

The new code can be detected in ``/u/land/fake_cvmfs/9.13/upexps``:
```
$  grep ../califa/califa.spec */*.spec   | sed s@/.*@@ | uniq
202202_para
202204_tests
202205_s509
202205_s522
califa_only
```

The old version can be detected using:
```
$ grep ^FEBEX3_CALIFA_BASE'()' */*.spec  | sed s@/.*@@ | uniq
201810_s444
201902_s444
201902_s473
201904_s454
201911_eng
201911_eng2
202002_s444
202002_s467
202004_s444
202006_s444
202006_test
202103_s455
202104_s515
202105_s494
califa2021 # false positive
califaKrakow17
califalisbon16
```
To get the overflow flags, we should convert the unpackers to use the new ``../califa/califa.spec`` instead. This will require new mapping files for all of them. (Technically, we could also fix the old version, but even that will require adding CALIFA_OV_%d to the mapping_CALIFA.hh)

Mapping files compatible with califa.spec can be generated from ``califa_mapping.txt`` by copying/cloning  ``/u/land/klenze/r3bmap/`` (or landgw01:git/r3bmap) and running the updated ``./run/califa/mapping/ucesb_mapping_gen_CALIFA_TOT.py``. 

This does require the original ``califa_mapping.txt``, which I do not generally have. (They are different for every experiment as they depend on the presumed gain settings.)

The replacement in the spec file should look like in s522:
```
...
#define CALIFA_DEFAULT_MAPPING 0
#include "../califa/califa.spec"
#include "mapping_CALIFA_s522.hh" // <--- this has to be generated per experiment
...
EVENT
{
...
        revisit califa_messel    = CALIFAM(type = 100, subtype = 10000, procid = 13, control = 90); // keep settings from old version
        revisit califa_wixhausen = CALIFAW(type = 100, subtype = 10000, procid = 13, control = 91); // same
...
}
...
```

[...]

PR #710, opened 2022-11-30
Remove '#ifdef EXT_DATA_ITEM_MAP_MATCH', to always use struct_map_success

inkdot7:
See discussion in #705 .

This patch would make sure to always check that all requested data members have sources in the input.

Please someone try it out, so we can figure out either why the code was not called, or why it does not report the missing overflow item.


hapol:
Checked at https://github.com/R3BRootGroup/R3BRoot/issues/705#issuecomment-1333861744

PR #712, opened 2022-12-05
Report FATAL error when ext_data_clnt::setup() failed to map all items.

inkdot7:
See discussion in #705 .

This patch to exit processing if any requested data members have no source in the input, such that user is alerted and does not get unexpected results.

Please try it out to see what detectors may have issues.

### DIFF
--- a/r3bsource/base/R3BUcesbSource.cxx
+++ b/r3bsource/base/R3BUcesbSource.cxx
@@ -173,9 +173,9 @@ Bool_t R3BUcesbSource::InitUnpackers()
         R3BLOG(warn, "ext_data_clnt::setup() failed to map all items:");
         ext_data_struct_info_print_map_success(fStructInfo, stderr, map_ok);
         /* FairRunOnline::Init() ignores the return value from
-         * GetSource()->InitUnpackers(); so do a (FATAL) error.
+         * GetSource()->InitUnpackers(); so do a FATAL error.
          */
-        R3BLOG(error,
+        R3BLOG(fatal,
                "ext_data_clnt::setup() mapping failure may "
                "cause unexpected analysis results due to missing "
                "data members. Unpacker needs fixing.");


e12exp:
In theory, we might want to have a way to whitelist items which are considered optional. 

In practice, all of the unpackers which do not use califa.spec seem to be affected by #705, so the need to support unpackers which do not use the per-hit califa wrts field is not there for now.

inkdot7:
@e12exp Do I understand correctly that (at least for CALIFA), all unpackers that would exit() with this patch would otherwise deliver bad data, such that the exit() ensures data quality by forcing update of the unpackers?

jose-luis-rs:
With the FATAL option we will have problems with the data processing of many experiments:

Example for the s515 experiment:
AMS:
SST5                                     49440    (0x04) not found
SST5I                                    49444    (0x04) not found
SST5E                                    53540    (0x04) not found
SST6                                     57636    (0x04) not found
SST6I                                    57640    (0x04) not found
SST6E                                    61736    (0x04) not found

NeuLAND:
NN_P25tcl_T1BM                           7579260  (0x04) not found
NN_P25tcl_T1BMI                          7579264  (0x04) not found
NN_P25tcl_T1BME                          7579464  (0x04) not found
NN_P25tcl_T1B                            7579664  (0x04) not found
NN_P25tcl_T1Bv                           7579668  (0x04) not found
NN_P25tcl_T2BM                           7585668  (0x04) not found
NN_P25tcl_T2BMI                          7585672  (0x04) not found
NN_P25tcl_T2BME                          7585872  (0x04) not found
NN_P25tcl_T2B                            7586072  (0x04) not found
NN_P25tcl_T2Bv                           7586076  (0x04) not found
NN_P25tfl_T1BM                           7592076  (0x04) not found
NN_P25tfl_T1BMI                          7592080  (0x04) not found
NN_P25tfl_T1BME                          7592280  (0x04) not found
NN_P25tfl_T1B                            7592480  (0x04) not found
NN_P25tfl_T1Bv                           7592484  (0x04) not found
NN_P25tfl_T2BM                           7598484  (0x04) not found
NN_P25tfl_T2BMI                          7598488  (0x04) not found
NN_P25tfl_T2BME                          7598688  (0x04) not found
NN_P25tfl_T2B                            7598888  (0x04) not found
NN_P25tfl_T2Bv                           7598892  (0x04) not found
...

MWPCs:
SOFMWPC2Plane1Q                          7689480  (0x04) not found
SOFMWPC2Plane1QI                         7689484  (0x04) not found
SOFMWPC2Plane1Qv                         7690764  (0x04) not found
SOFMWPC2Plane2Q                          7692044  (0x04) not found
SOFMWPC2Plane2QI                         7692048  (0x04) not found
SOFMWPC2Plane2Qv                         7693328  (0x04) not found
SOFMWPC2Plane3Q                          7694608  (0x04) not found
SOFMWPC2Plane3QI                         7694612  (0x04) not found
SOFMWPC2Plane3Qv                         7695892  (0x04) not found
SOFMWPC3Plane1Q                          7697172  (0x04) not found
SOFMWPC3Plane1QI                         7697176  (0x04) not found
SOFMWPC3Plane1Qv                         7698456  (0x04) not found
SOFMWPC3Plane2Q                          7699736  (0x04) not found
SOFMWPC3Plane2QI                         7699740  (0x04) not found
SOFMWPC3Plane2Qv                         7701020  (0x04) not found
SOFMWPC3Plane3Q                          7702300  (0x04) not found
SOFMWPC3Plane3QI                         7702304  (0x04) not found
SOFMWPC3Plane3Qv                         7703584  (0x04) not found
SOFMWPC4Plane1Q                          7704864  (0x04) not found
SOFMWPC4Plane1QI                         7704868  (0x04) not found
SOFMWPC4Plane1Qv                         7706148  (0x04) not found
SOFMWPC4Plane2Q                          7707428  (0x04) not found
SOFMWPC4Plane2QI                         7707432  (0x04) not found
SOFMWPC4Plane2Qv                         7708712  (0x04) not found
SOFMWPC4Plane3Q                          7709992  (0x04) not found
SOFMWPC4Plane3QI                         7709996  (0x04) not found
SOFMWPC4Plane3Qv                         7711276  (0x04) not found

Example for the s455 experiment (part 1):

[WARN] R3BUcesbSource.cxx:173:InitUnpackers(): ext_data_clnt::setup() failed
Item                                     Offset   Success
---------------------------------------- -------- ---------------------------
TPATFAIL                                 692      (0x04) not found
SOFSCI2TFM                               1993020  (0x04) not found
SOFSCI2TFMI                              1993024  (0x04) not found
SOFSCI2TFME                              1993036  (0x04) not found
SOFSCI2TF                                1993048  (0x04) not found
SOFSCI2TFv                               1993052  (0x04) not found
SOFSCI2TCM                               1994252  (0x04) not found
SOFSCI2TCMI                              1994256  (0x04) not found
SOFSCI2TCME                              1994268  (0x04) not found
SOFSCI2TC                                1994280  (0x04) not found
SOFSCI2TCv                               1994284  (0x04) not found
SOFSCI3TFM                               1995484  (0x04) not found
SOFSCI3TFMI                              1995488  (0x04) not found
SOFSCI3TFME                              1995500  (0x04) not found
SOFSCI3TF                                1995512  (0x04) not found
SOFSCI3TFv                               1995516  (0x04) not found
SOFSCI3TCM                               1996716  (0x04) not found
SOFSCI3TCMI                              1996720  (0x04) not found
SOFSCI3TCME                              1996732  (0x04) not found
SOFSCI3TC                                1996744  (0x04) not found
SOFSCI3TCv                               1996748  (0x04) not found
SOFSCI4TFM                               1997948  (0x04) not found
SOFSCI4TFMI                              1997952  (0x04) not found
SOFSCI4TFME                              1997964  (0x04) not found
SOFSCI4TF                                1997976  (0x04) not found
SOFSCI4TFv                               1997980  (0x04) not found
SOFSCI4TCM                               1999180  (0x04) not found
SOFSCI4TCMI                              1999184  (0x04) not found
SOFSCI4TCME                              1999196  (0x04) not found
SOFSCI4TC                                1999208  (0x04) not found
SOFSCI4TCv                               1999212  (0x04) not found
SOFTOFW_P1E1                             2056236  (0x04) not found
SOFTOFW_P1E2                             2056240  (0x04) not found
SOFTOFW_P2E1                             2056452  (0x04) not found
SOFTOFW_P2E2                             2056456  (0x04) not found
SOFTOFW_P3E1                             2056668  (0x04) not found
SOFTOFW_P3E2                             2056672  (0x04) not found
SOFTOFW_P4E1                             2056884  (0x04) not found
SOFTOFW_P4E2                             2056888  (0x04) not found
SOFTOFW_P5E1                             2057100  (0x04) not found
SOFTOFW_P5E2                             2057104  (0x04) not found
SOFTOFW_P6E1                             2057316  (0x04) not found
SOFTOFW_P6E2                             2057320  (0x04) not found
SOFTOFW_P7E1                             2057532  (0x04) not found
SOFTOFW_P7E2                             2057536  (0x04) not found
SOFTOFW_P8E1                             2057748  (0x04) not found
SOFTOFW_P8E2                             2057752  (0x04) not found
SOFTOFW_P9E1                             2057964  (0x04) not found
SOFTOFW_P9E2                             2057968  (0x04) not found
SOFTOFW_P10E1                            2058180  (0x04) not found
SOFTOFW_P10E2                            2058184  (0x04) not found
SOFTOFW_P11E1                            2058396  (0x04) not found
SOFTOFW_P11E2                            2058400  (0x04) not found
SOFTOFW_P12E1                            2058612  (0x04) not found
SOFTOFW_P12E2                            2058616  (0x04) not found
SOFTOFW_P13E1                            2058828  (0x04) not found
SOFTOFW_P13E2                            2058832  (0x04) not found
SOFTOFW_P14E1                            2059044  (0x04) not found
SOFTOFW_P14E2                            2059048  (0x04) not found
SOFTOFW_P15E1                            2059260  (0x04) not found
SOFTOFW_P15E2                            2059264  (0x04) not found
SOFTOFW_P16E1                            2059476  (0x04) not found
SOFTOFW_P16E2                            2059480  (0x04) not found
SOFTOFW_P17E1                            2059692  (0x04) not found
SOFTOFW_P17E2                            2059696  (0x04) not found
SOFTOFW_P18E1                            2059908  (0x04) not found
SOFTOFW_P18E2                            2059912  (0x04) not found
SOFTOFW_P19E1                            2060124  (0x04) not found
SOFTOFW_P19E2                            2060128  (0x04) not found
SOFTOFW_P20E1                            2060340  (0x04) not found
SOFTOFW_P20E2                            2060344  (0x04) not found
SOFTOFW_P21E1                            2060556  (0x04) not found
SOFTOFW_P21E2                            2060560  (0x04) not found
SOFTOFW_P22E1                            2060772  (0x04) not found
SOFTOFW_P22E2                            2060776  (0x04) not found
SOFTOFW_P23E1                            2060988  (0x04) not found
SOFTOFW_P23E2                            2060992  (0x04) not found
SOFTOFW_P24E1                            2061204  (0x04) not found
SOFTOFW_P24E2                            2061208  (0x04) not found
SOFTOFW_P25E1                            2061420  (0x04) not found
SOFTOFW_P25E2                            2061424  (0x04) not found
SOFTOFW_P26E1                            2061636  (0x04) not found
SOFTOFW_P26E2                            2061640  (0x04) not found
SOFTOFW_P27E1                            2061852  (0x04) not found
SOFTOFW_P27E2                            2061856  (0x04) not found
SOFTOFW_P28E1                            2062068  (0x04) not found
SOFTOFW_P28E2                            2062072  (0x04) not found

We have general data structures for the SOFIA detectors that contain:
SOFSCI --> 4 scintillators
SOFMWPC --> 4 mwpcs
SOFTOFW where we have the deposited energy in some cases

Summary per experiment:
- S455 part 1: 1 sofsci, 4 mwpcs, 6 AMSs
- S455 part 2: 2 sofsci, 4 mwpcs
- S467: 4 sofsci, 3-4 mwpcs
- S509 and S522: 1 sofsci, 2 mwpcs 
- S515: 1 sofsci, 1 mwpc, 4 AMSs

klenze:
> @e12exp Do I understand correctly that (at least for CALIFA), all unpackers that would exit() with this patch would otherwise deliver bad data, such that the exit() ensures data quality by forcing update of the unpackers?

Philipp aka @klenze aka @e12exp here. CALIFA_OV is only defined for s509 and s522. All the others should be considered broken. 

The case where just the WRTS per channel are missing does not appear.

inkdot7:
I believe some list of patterns (e.g. "SOFSCI4.*") unpacker items that are *expected* to be missing needs to be filled from either the code which sets up the structures, or some higher-level using code (whichever is most convenient), for the various experiments.  Of course, if such items are found, it is then also an error, as expectations need to be corrected.

ryotani:
I would like to move the conversation occurring in #705 here. In the analysis WG meeting, this issue was discussed. Although we recognise the importance of making unpacker and R3BRoot consistent, practically it is not easy as the detector setup evolves time-to-time. If we would like to keep compatibility for the old experiment, it is often that we don't have the same number of channels that we have right now. Of course, it is highly encouraged to sort these issues as much as possible, can it be WARNING instead of FATAL, or set a variable accessible from outside of the class to ignore the mismatch?
Or if there's still a way to avoid these warnings, could anyone instruct me/us on how can we solve the issue? I presume someone from the upexps side would be required to create ext_h101 files for each experimental configuration.

jose-luis-rs:
@ryotani the best option could be to create data structures with mandatory and optional variables. In the case of SOFSCI one scintiallator is always mandatory, the others can be optional. For mwpcs, mwpc0 is mandatory, the others are optional. 

Maybe upexps experts have more ideas today ...

inkdot7:
It should be is possible to handle changing number of channels between experiments with one code by *declaring* the data structures either with varying size (likely not nice) or by letting them have the largest size needed (for any experiment).  Then only *request filling* of the size (number of array members) needed by the experiment, i.e. the same size as the relevant unpacker provides.

To do this, one should not just copy generate `ext_h101` files, but use (one of) them as a starting point for a detector (or a structure) and edit it to make the variable sizes given by some variables.

The `ext_data` interface provides functions for declaring the destination structure.  `ext_data_struct_info_alloc()`, `ext_data_struct_info_item()`, which are used by the macros `EXT_STR_ITEM_INFO...` which in turn are used in the generated `EXT_STR_something_ITEMS_INFO`.

For further code reduction, when several blocks of same variables are needed, the amount of code code could be reduced by putting some loop around it and then having the names generated in parts.

inkdot7:
Just a heads up: fairroot has merged their bugfix (https://github.com/FairRootGroup/FairRoot/pull/1307) two weeks ago.    This fix will then not be needed, but it would probably be good to have in the meantime, such that r3broot is kept clean of these kinds of unpacking errors.

A way to fix the mappings was suggested in my last comment above dated Dec 15.

jose-luis-rs:
> Just a heads up: fairroot has merged their bugfix ([FairRootGroup/FairRoot#1307](https://github.com/FairRootGroup/FairRoot/pull/1307)) two weeks ago. _This_ PR will then not be needed, but it would probably be good to have in the meantime, such that r3broot is kept clean of these kinds of unpacking errors.
> 
> A way to fix the mappings was suggested in my last comment above dated Dec 15.

I guess @YanzhaoW will comment on it soon

YanzhaoW:
@inkdot7 @jose-luis-rs We are currently using FairRoot 18.8 patch both in CI and lxir136. In the CI, it's directly sourced from cvmfs in a gsi server. I will look into it if we could get the update from that patch.

inkdot7:
Merging this PR will make users aware of __*very severe*__ unpacking errors and it has now been around for 6 months.

It passes CI.

Could we now merge this, such that users become aware of related unpacking issues and thus are urgently encouraged to solve them.

This is a direct request from the UCESB author and maintainer.  UCESB does not want to be associated with erroneous unpacking due to lacking checks.

jose-luis-rs:
> Merging this PR will make users aware of **_very severe_** unpacking errors and it has now been around for 6 months.
> 
> It passes CI.
> 
> Could we now merge this, such that users become aware of related unpacking issues and thus are urgently encouraged to solve them.
> 
> This is a direct request from the UCESB author and maintainer. UCESB does not want to be associated with erroneous unpacking due to lacking checks.

In #862 I have included a new repo to check the unpacking, but we need access to upexps to ensure that everything works.
How to do this?

inkdot7:
upexps is here (https://git.gsi.de/r3b/upexps), but not public.  I do not know if it has been scrutinized to be made publicly available.  Its CI does not pass (does not even start), so general status unknown.  @bl0x , @hanstt may have more info.

klenze:
One thing we should think about is how to handle less-than-maximum-length conditions. 

For CALIFA, we store the data in one-dimensional structures, which fail gracefully if the array length provided by the unpacker is shorter. 

For detectors which use onionized multi-dimensional arrays which get mapped to multiple 1d fields, things are more precarious?

The fact that complaints about SOFTOFW_P13E2 appeared likely means that if there was some experiment where just P1 to P12 were in use, this would actually blow up?

What is a good approach here? Make every slice except for the first optional? Have the caller state how many PSPs (or TOF or Neuland planes) they expect to be present, and assert we have exactly as many?

\begin{musing}

In the long run, I guess there could be a better system to handle what is presently done by the auto-generated EXT_STR_h101_FOO_ITEMS_INFO. I am still unsure what the best way would be, though. Reflection is not really a strong point of C++. ~~ In theory, you could have a macro which defines a single entry struct with a constructor handling the setup, and then multi-inheriting from all the classes defined by macro calls ~~ _actually you can not, because the spoilsports at ISO do not allow to either define a nameless class nor use a class definition instead of a class name in either a template parameter (pack) or an inheritance specification_. 

So, due to inherent limitations of the template mechanism, we can't do this. The only part of the language which can stringify arguments is the preprocessor. If we really want to avoid spelling out all the variables twice we could write a higher order macro. [Loops](https://www.boost.org/doc/libs/1_81_0/libs/preprocessor/doc/index.html) in the preprocessor are possible but somewhat scary. We would be better of autogenerating the setup code (with loops over onion structs and everything) in python. 

\end{musing}

Also, I think this PR should be merged. (Possibly after CI is including recent unpackers, if that can be merged quickly.)

inkdot7:
> The fact that complaints about SOFTOFW_P13E2 appeared likely means that if there was some experiment where just P1 to P12 were in use, this would actually blow up?

Indeed, that would blow up in the post-setup check of `struct_map_success` which will have accumulated such missing item issues.   Either (not recommended) one would have to go through all the items with `ext_data_struct_info_map_success()` and figure out which misses are acceptable, or (recommended) the mapping request (from `EXT_STR_ITEM_INFO` et al.) needs to only request what it knows is available.

When only asking mapping of fewer items than are available in the data structures, R3BRoot must then not look at the contents of the further (non-requested) data items, or it might copy non-filled data!  That is a problem outside the scope of the external data interface (since it does not know about those members), but one which needs to be carefully thought about!

> What is a good approach here? Make every slice except for the first optional? Have the caller state how many PSPs (or TOF or Neuland planes) they expect to be present, and assert we have exactly as many?

It sounds reasonable to require the first slice, since if that does not exist, one should not have asked for that data structure at all.

I would also suggest to loopify (manually!) the calls to `EXT_STR_ITEM_INFO` and generate the strings with changing indices from the loop variable.  That would reduce the amount of code to review.

Carthago delenda est:

Please consider merging this PR now! - avoid bad unpacking by mistake should be higher priority than disruptions.

inkdot7:
In ucesb/ext_data interface, `EXT_STR_ITEM_INFO...` has been renamed `EXT_STR_ITEM_INFO2...` to take an additional `flags` argument.  If 0 then behaves as before, and that is what is generated.  If `EXT_DATA_ITEM_FLAGS_OPTIONAL` is given, then the mapping will report missing items as `EXT_DATA_ITEM_MAP_OPT_NOT_FOUND` instead of `EXT_DATA_ITEM_MAP_NOT_FOUND` and this is included in `EXT_DATA_ITEM_MAP_OK` so will not generate an error during after-setup check.

I.e. `EXT_DATA_ITEM_FLAGS_OPTIONAL` can be used to mark items as optional.  It is up to the using code to make sure they are set to good default values, such that it can handle when they are not assigned by `ext_data_fetch_event()`.

Carthago delenda est:

Please consider merging this PR now! - avoid bad unpacking by mistake should be higher priority than disruptions.

jose-luis-rs:
This PR requires the dev branch of FairRoot, is everything ready at GSI to manage the unpacking?

inkdot7:
> This PR requires the dev branch of FairRoot, is everything ready at GSI to manage the unpacking?

As far as I know, this patch does *not* require the dev branch of FairRoot.  (As shown by passing CI?)

It rather alerts to mapping errors without having the newer FairRoot.  However: newer FairRoot versions will by themselves report fatal, since they now care about the false return from `R3BUcesbSource::InitUnpackers()`.  That was the FairRoot bugfix.

jose-luis-rs:
Ok, but at the moment we don't have CIs to check the unpacking process!

inkdot7:
> Ok, but at the moment we don't have CIs to check the unpacking process!

At the moment, I get reports about bad data being mapped due to this missing check.  Yes - this will break things for people with broken mappings.  But that must surely be better than to continue analyse with unknown data?

inkdot7:
And...  any affected users surely can do a local modification that changes the `R3BLOG(fatal,` back to `R3BLOG(error,`.  But then they have acknowledged that they have been warned about the unpacking issues.

jose-luis-rs:
Yes, so I will merge it now.

ryotani:
Hi, I'm working on the unpacker, 202002_s467 to eliminate the issues for
undefined mappings. Although most of them could be solved, I could not
solve some values for CALIFA.
```
CALIFA_TRGENE                            48788    (0x04) not found
CALIFA_TRGENEI                           48792    (0x04) not found
CALIFA_TRGENEv                           48800    (0x04) not found
```
Apparently, they have been defined in the PR #591 by @jose-luis-rs.
However, I cannot find any description of these variables in unpacker in
the gsi server. I think this will affect everyone who's using
CalifaFebexReder. If you could instruct us on how to define them, that
would be great. @jose-luis-rs @klenze, do you know anything about this?

@jose-luis-rs I have another question concerning in the commit you made in
SOFIA repo:
https://github.com/R3BRootGroup/sofia/commit/c6d55c4fa6ffd8640cd68d238a95722a9410d201#diff-9516b3f0736fb252dbb639a396cbbc71d7134738dcc5a0cb3efd9d48ef3f1adfL55-R99
You changed the name of TIMESTAMP of SOFIA from TIMESTAMP_SOFIA_WR_T1 to
TIMESTAMP_SOFIA1WR_T1. As far as I can see in the gsi server, the unpackers
using SOFIA detectors, such as s455, define it as TIMESTAMP_SOFIA_1_WR_T1.
Do you think it's problematic to change the R3B side to adapt to the other
unpackers?

klenze:
> CALIFA_TRGENE

As far as I know, this is the length of the T1 trigger from main DAQ which we use for validation, as measured by FEBEX (as an amplitude). 
TRLO II uses pulse width modulation to encode one of eight states into each trigger, so that we can correlate that info for different detector systems. This allows us to detect the case where events are marked with a previous white rabbit time stamps, as the PSPs were in some previous experiment. 

The CALIFA DAQ is monitored using go4, where we bypass ucesb, so I did not put these "special channels" into any mapping. 

To my knowledge, @jose-luis-rs added the mapping. I assume he did so because simply adding them as channels (perhaps crystal ids 0xffff and 0xfffe) would cause spikes in the histograms of all mapped channels. However, "all mapped channels" is not a terribly useful thing to plot anyhow, for energies, you would mix both gain ranges, for timing differences, you would mix both sides. My recommendation would be to encode the special channels with special crystal ids, and register that crystal id as being special.

jose-luis-rs:
> Hi, I'm working on the unpacker, 202002_s467 to eliminate the issues for undefined mappings. Although most of them could be solved, I could not solve some values for CALIFA. ``` CALIFA_TRGENE 48788 (0x04) not found CALIFA_TRGENEI 48792 (0x04) not found CALIFA_TRGENEv 48800 (0x04) not found ``` Apparently, they have been defined in the PR #591 by @jose-luis-rs. However, I cannot find any description of these variables in unpacker in the gsi server. I think this will affect everyone who's using CalifaFebexReder. If you could instruct us on how to define them, that would be great. @jose-luis-rs @klenze, do you know anything about this? 

These special channels were implemented in 2021 after the fission experiments(s455), so in your case you cannot define them. Sorry for that.

@jose-luis-rs I have another question concerning in the commit you made in SOFIA repo: [R3BRootGroup/sofia@c6d55c4#diff-9516b3f0736fb252dbb639a396cbbc71d7134738dcc5a0cb3efd9d48ef3f1adfL55-R99](https://github.com/R3BRootGroup/sofia/commit/c6d55c4fa6ffd8640cd68d238a95722a9410d201#diff-9516b3f0736fb252dbb639a396cbbc71d7134738dcc5a0cb3efd9d48ef3f1adfL55-R99) You changed the name of TIMESTAMP of SOFIA from TIMESTAMP_SOFIA_WR_T1 to TIMESTAMP_SOFIA1WR_T1. As far as I can see in the gsi server, the unpackers using SOFIA detectors, such as s455, define it as TIMESTAMP_SOFIA_1_WR_T1. Do you think it's problematic to change the R3B side to adapt to the other unpackers?

What do you mean with "R3B side to adapt to the other unpackers"?
which unpackers?

ryotani:
Hi @klenze and @jose-luis-rs, thanks for the information. I modified the upexps and seems to work fine.
In the preamble of mapping_CALIFA_s467.hh, you @klenze mentioned that it was auto-generated by ./run/califa/mapping/ucesb_mapping_gen_CALIFA_TOT.py, but I just added several lines manually. Hope it will not cause a problem.
I added these lines:
```
SIGNAL(ZERO_SUPPRESS:CALIFA_TRGENE_1);
SIGNAL(CALIFA_TRGENE_1,califa_messel.febex3.gosip.hit.base.energy[1028],DATA16);
SIGNAL(CALIFA_TRGENE_2,califa_wixhausen.febex3.gosip.hit.base.energy[1028],DATA16);
```

I noticed the timestamp for sofia DAQ is defined differently in unpackers (such as s455) and the R3BSofWhiterabbitReader.cxx. The former names variables as TIMESTAMP_SOFIA_1_WR_T1, while the latter names TIMESTAMP_SOFIA1WR_T1.

An example case of the unpacker can be found:
```
land@lxir136:202104_s455 > pwd
/u/land/fake_cvmfs/10/cvmfs_fairroot_v18.8.0_fs_nov22p1_extra/upexps/202104_s455
land@lxir136:202104_s455 > grep TIMESTAMP_SOFIA ./*
Binary file ./202104_s455 matches
Binary file ./202104_s455_part2 matches
grep: ./gen_202104_s455: Is a directory
grep: ./gen_202104_s455_part2: Is a directory
grep: ./mapping_mwpc: Is a directory
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_1_ID, sofia_se_ts.ts.subsystem_id, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_1_WR_T1, sofia_se_ts.ts.t1, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_1_WR_T2, sofia_se_ts.ts.t2, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_1_WR_T3, sofia_se_ts.ts.t3, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_1_WR_T4, sofia_se_ts.ts.t4, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_2_ID, sofia_me_ts.ts.subsystem_id, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_2_WR_T1, sofia_me_ts.ts.t1, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_2_WR_T2, sofia_me_ts.ts.t2, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_2_WR_T3, sofia_me_ts.ts.t3, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIA_2_WR_T4, sofia_me_ts.ts.t4, DATA16);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIAM_WR_T1, sofia_mwpc.data.ts.time_lo, DATA32);
./mapping_timestamp.hh:SIGNAL(TIMESTAMP_SOFIAM_WR_T2, sofia_mwpc.data.ts.time_hi, DATA32);
```

ryotani:
Okay, I was informed that the names of variables are not recommended to have unnecessary underscores in ucesb. Then it makes sense. So ignore the latter comment in the previous post.
`It is not recommended to use digits or underscores as part of variable names.` in the USESB documentation.

YanzhaoW:
Hi @inkdot7 

As the current code:
https://github.com/R3BRootGroup/R3BRoot/blob/29ff3c654d4f8461ab0580765295ef176bff1082/r3bsource/base/R3BUcesbSource.cxx#L170-L183

If we allow some client side items unmatched to the server side items, shouldn't we allow the flag `EXT_DATA_ITEM_MAP_NOT_FOUND` with this change:

```cpp
    uint32_t map_ok = EXT_DATA_ITEM_MAP_OK | EXT_DATA_ITEM_MAP_NO_DEST | EXT_DATA_ITEM_MAP_NOT_FOUND;
```

inkdot7:
> If we allow some client side items unmatched to the server side items, shouldn't we allow the flag `EXT_DATA_ITEM_MAP_NOT_FOUND` with this change:
> 
> ```c++
>     uint32_t map_ok = EXT_DATA_ITEM_MAP_OK | EXT_DATA_ITEM_MAP_NO_DEST | EXT_DATA_ITEM_MAP_NOT_FOUND;
> ```

I think that would undo the entire purpose of this commit - to make sure items are not missing.  If items may be missing sometimes, then the client needs to check what was mapped and what was not mapped before use.

YanzhaoW:
@inkdot7 

I did a test where I use the ext_h101 file generated from the s522 as the client and unpacker from s444 as the server. As you have recently suggested, we should't copy and paste the ext header for each experiment to R3BRoot and instead we should just use one ext headers for as many experiments as possible. In s522, we had 26 planes and triggers are also implemented as well. In s444, we have 16 planes and no trigger.

The "array short" flag is expected as we had shortened the array size from 10000 to 1500. However, we get "not found" flags for the signals from plane 17 to plane 26. Our current code forces the program to abort if either "not found" or "array short" is found in any client items. I may misunderstand your suggestion.

Or do you actually mean we use the executable from s522 to unpack the data from s444 (both on the server side)? But if so, ext headers are kind of irrelevant.

inkdot7:
The suggestion was to only request those items (here planes) which are expected.  So each experiment just needs to know how many planes to ask for, and then ask for those items only.  That should be most easy with a loop and having made the structure such that e.g. each plane is the structure being asked for.  Just like the different detectors are treated separately, one can subdivide further within a detector too.  When setting up the request, the plane numbers need to change, so some `sprintf`ing to take care of that.  The data structure itself (that is allocated in r3broot) can then be large enough to handle the largest case.

YanzhaoW:
Ok, I guess this needs also to be done in runtime. I'm not sure what's the best way to implement this "items that are only asked for". For now, what comes to my mind is something like this:

```cpp
    auto struct_map_success = uint32_t{0};
    auto status = fClient.setup(NULL, 0, &fStructInfo, &struct_map_success, fEventSize);
    if (!status)
    {
        R3BLOG(fatal, "UCESB error: " << fClient.last_error());
        return kFALSE;
    }
    auto map_ok = EXT_DATA_ITEM_MAP_OK | EXT_DATA_ITEM_MAP_NO_DEST | EXT_DATA_ITEM_MAP_IGNORE;
    
    for(const auto& reader : readers)
    {
          reader->SetStructInfoIgnore(fStructInfo);
    }
    auto is_map_ok = CheckIfMapOk(fStructInfo,  map_ok);
    if(!is_map_ok) { R3BLOG(fatal) << "";}
```
This is just implementation only on R3BRoot side. But in this case `struct_map_success` is not used at all.
I would say it would be much better if we could implement this both in R3BRoot and in ucesb client side. If so, we could first set the ignore flag before calling the setup function. And in the setup function (ucesb side), we don't do the matching if the item has ignore flag. In the end, we could allow struct_map_success have the ignore flag.

klenze:
@inkdot7: please do not mention the existence of [sprintf](http://natashenka.ca/wp-content/uploads/2014/01/sprintf8x11.png). 

I see two ways forward which I would consider an acceptable solutions (e.g. automatic generation of the h101 preprocessor calls). 

Given that we already allow the innermost array dimension to be short, I think that we could use a similar approach for the other dimensions. e.g. NeulandReader could require the first plane (or more specifically the first bar of the first plane) to be defined, and everything else is just treated like a short array. 

One solution would be to look into static reflection libraries for C++. The other solution would be to use an external tool to parse C or C++ (e.g. the onion struct) and autocreate the h101 setup from it. I think either would be an adequate programming challenge. (Bonus points for (a) supporting templated structs like ``template <int nPlanes, int nBarsPerPlane> ext_h101_neuland``, (b) allowing the struct to specify what should be optional, (c) solving the problem without writing out cascaded for loops.)

Of course, while overhauling the readers, other stuff to fix would be:
(a) auto-allocate the overall h101 object, instead of having the user specifying it explicitly in the macro
(b) support prefixes for both the h101 and output structure names. 

One could also debate if ucesb is really the right place for doing the mapping. I would argue that for some purposes, it is impractical to have a full mapping in the ucesb stage. For TAMEX, my understanding is that sixteen channels on a card share one trigger channel, so to abstract away all of the electronics mapping layer would require to duplicate the trigger channels sixteen-fold. (Of course, the practical alternative to doing the mapping in ucesb is storing the mapping in a FairParameter container within a root file. I do not advocate that.)

inkdot7:
I had forgotten that the interface to setup the structure `ext_data_struct_info_item()` actually can take a last argument flag, `EXT_DATA_ITEM_FLAGS_OPTIONAL`.  That marks an item as optional beforehand and then change the contribution to the map success return value for those.

However, this I think is not appropriate to use in this case.  When e.g. another plane in neuland is expected, its variables need to fully exist.  Not only some of them.  Letting such things pass would lead to the same problem that was in CALIFA, only in more subtle ways.

Thus the suggestion that R3Broot e.g. in the neuland case knows how many planes it should get data for, and it makes exactly the calls to `ext_data_struct_info_item()` that are needed for that.  That does not require the parsing of a structure, that just requires one number, which even could be runtime defined.

If there are other things that exist for some experiments, and do not for others, that r3broot can also know by some further configuration setting, and then ask or not ask for those items.

It would of course be nicer to be able to query what is available before setting up the request, but for the moment that is unfortunately not possible.  Or is it?  Looks like `ext_data_setup()` might even be prepared such that it can be called many times.  If so, likely not (well) tested, and it should not actually set structures up for just doing queries, so some more handling would be needed.  So for the moment, just knowing e.g. the number of planes is more straightforward.

Also, doing the mapping already in ucesb might not be the better choice.  But r3broot can always choose to ask for the data (also) at an earlier unpack levels, or even read the lmd files itself.

PR #715, opened 2022-12-06
Small spell fix that triggers format check failure.

inkdot7:
This commit changes a file which also has a clang-format error, and thus triggers to CI error.  That is intentional to see that the format check works.

It is based on #714 which re-enables the format checking.

After #714 has been applied, I can fix the formatting of this commit such that it passes CI before being merged.

### DIFF
--- a/r3bsource/base/R3BUcesbSource.cxx
+++ b/r3bsource/base/R3BUcesbSource.cxx
@@ -73,7 +73,7 @@ Bool_t R3BUcesbSource::Init()
     else
     {
         eventHeader = new R3BEventHeader();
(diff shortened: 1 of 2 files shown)


jose-luis-rs:
Done!

inkdot7:
Let's see if this triggers the fault, and then fix it...

inkdot7:
But that would require to re-open this PR.  I have no button for that.

jose-luis-rs:
Well, it worked with my current PR #706 so I guess that we will not need to re-open it.

For a strange reason I am not allowed to re-open it

Issue #716, opened 2022-12-07
Please enable -Werror in CI builds


inkdot7:
Looking around the CI system, this is found (scroll to the last warning):

https://cdash.gsi.de/viewBuildError.php?type=1&buildid=359806

which references current (merged) code that indeed looks suspicious.  Either it should be a single `&`, or the `<<` is not needed:

https://github.com/R3BRootGroup/R3BRoot/blob/4e2decfab13d80ee412c6dfeabfcc19c4efd4976/analysis/online/R3BOnlineSpillAnalysis.cxx#L365

having the CI find such issues should be much cheaper than debugging them by hand later.

klenze:
I am working on it in #718. 

At the moment, clang-format-8 still complains in CI. 
As clang-format-8 is not installed on lxir136, I will try to fix the formatting locally with clang-format-9.

jose-luis-rs:
I think that we can close this issue.

inkdot7:
I'm not able to follow how the CI jobs are run or what flags they set.  I assume the merged commit above solves the issue.

jose-luis-rs:
It is made from the CMakeLists file, [here](
https://github.com/R3BRootGroup/R3BRoot/blob/e5e975dd601dc07330b1a35bc0df36fe5d8ee269/CMakeLists.txt#L223)

PR #718, opened 2022-12-07
Fixed warnings found with gcc-10. Enabled Werror for CI builds.

klenze:


klenze:
Should work now, fixes #716.

klenze:
Motivation for this PR: People sometimes commit code which creates new warnings. About half the time, the warnings are actually due to bugs (such as && being used instead of &) or risky behavior (such as not initializing a pointer (with nullptr) which is supposed to be set in some loop). 

I am very convinced that we should have one test case for compilation with -Werror on at least one platform. It could be a separate test case perhaps. (If we compile with make instead of make -k and use -Wfatal-errors, and just print out the last 50 lines of the compilation output, finding the offending code would be easier.)

If the proposal #716 is rejected, I expect an alternative on how errors which are visible as warnings can reliably be caught and fixed.

klenze:
> I think that the option "make" + "-Wfatal-errors" + "print out the last 50 lines" could be very useful.

Agreed. I will update my PR tomorrow. 

Ideally, we would also upload the full output to some web server so the user also gets useful feedback in case of very long warning messages (which may or may not exist. The error messages for templates are legendarily long).

jose-luis-rs:
Hi

The file travis-generate-dockerfile.sh was deleted but this is still used by the hidden script ".travis.yml". Please, comment the line:
 - ${TRAVIS_BUILD_DIR}/travis-generate-dockerfile.sh | tee Dockerfile

Then I will see if we could add some specific tests to check other repositories like sofia, asyeos, glad, etc. This will help us to be sure everything is working properly.

Thanks in advance

klenze:
see #729 for a sample failure. If anyone has a simpler way to output just the error, I am open to it. 

Also, I am kind of confused by the relationship between .travis.yml and .github/workflows/main.yml. 

From my understanding, the github CI replaced travis and the travis stuff is obsolete except for the travis-clang-format-check.sh which is called from the github CI. Am I wrong there?

jose-luis-rs:
Hi, let me know when you finish with this PR to merge it. Thanks in advance

klenze:
Hi,
I would say it is ready.
Cheers, 
   Philipp

jose-luis-rs:
Ok, please, could you rebase the R3BRoot/dev branch?

PR #720, opened 2022-12-11
Renamed CalifaHit to CalifaCluster

klenze:
The name "R3BCalifaHitData" is a relic from ancient times, much like the class name "R3BCalifa". 

Nice person that I am, I added using declarations so that the old name can still be used. 

(Yes, this breaks from the Mapped -> Cal (or for whatever reason, CrystalCal) -> Hit scheme. This is intentional. The fixed "there are exactly two steps in data processing for any detector" is just not true for many detectors.

### DIFF
--- a/analysis/online/R3BCalifavsFootOnlineSpectra.cxx
+++ b/analysis/online/R3BCalifavsFootOnlineSpectra.cxx
@@ -12,8 +12,8 @@
  ******************************************************************************/
 
 #include "R3BCalifavsFootOnlineSpectra.h"
+#include "R3BCalifaClusterData.h"
 #include "R3BCalifaCrystalCalData.h"
-#include "R3BCalifaHitData.h"
 #include "R3BEventHeader.h"
 #include "R3BFootCalData.h"
 #include "R3BFootHitData.h"
@@ -191,7 +191,7 @@ void R3BCalifavsFootOnlineSpectra::Exec(Option_t* option)
 
     for (Int_t ihit = 0; ihit < nHits; ihit++)
     {
-        auto hit = (R3BCalifaHitData*)fHitItemsCalifa->At(ihit);
+        auto hit = (R3BCalifaClusterData*)fHitItemsCalifa->At(ihit);
         if (hit->GetEnergy() < 50e3) // 50MeV
             continue;
 
--- a/analysis/online/R3BCalifavsTofDOnlineSpectra.cxx
+++ b/analysis/online/R3BCalifavsTofDOnlineSpectra.cxx
@@ -12,7 +12,7 @@
  ******************************************************************************/
 
 #include "R3BCalifavsTofDOnlineSpectra.h"
-#include "R3BCalifaHitData.h"
+#include "R3BCalifaClusterData.h"
 #include "R3BEventHeader.h"
 #include "R3BLogger.h"
 #include "R3BTofdHitData.h"
@@ -177,7 +177,7 @@ void R3BCalifavsTofDOnlineSpectra::Exec(Option_t* option)
     Int_t nHits = fHitItemsCalifa->GetEntriesFast();
     for (Int_t ihit = 0; ihit < nHits; ihit++)
     {
-        auto hit = (R3BCalifaHitData*)fHitItemsCalifa->At(ihit);
+        auto hit = (R3BCalifaClusterData*)fHitItemsCalifa->At(ihit);
         if (hit->GetEnergy() < fMinProtonE)
             continue;
 
@@ -198,7 +198,7 @@ void R3BCalifavsTofDOnlineSpectra::Exec(Option_t* option)
         Double_t califa_e[nHits];
         for (Int_t ihit = 0; ihit < nHits; ihit++)
         {
-            auto hit = (R3BCalifaHitData*)fHitItemsCalifa->At(ihit);
+            auto hit = (R3BCalifaClusterData*)fHitItemsCalifa->At(ihit);
             if (!hit)
                 continue;
             theta = hit->GetTheta() * TMath::RadToDeg();
--- a/califa/CMakeLists.txt
+++ b/califa/CMakeLists.txt
@@ -62,7 +62,7 @@ set(SRCS
 ./calibration/R3BCalifaMapped2CrystalCalPar.cxx
 ./calibration/R3BCalifaCrystalCal2TotCalPar.cxx
 ./calibration/R3BCalifaMapped2CrystalCal.cxx
-./calibration/R3BCalifaCrystalCal2Hit.cxx
+./calibration/R3BCalifaCrystalCal2Cluster.cxx
 ./calibration/R3BCalifaCrystalCal2CrystalCalPID.cxx
 ./online/R3BCalifaOnlineSpectra.cxx
 ./online/R3BCalifaJulichOnlineSpectra.cxx
--- a/califa/CalifaLinkDef.h
+++ b/califa/CalifaLinkDef.h
@@ -33,7 +33,7 @@
 #pragma link C++ class R3BCalifaMapped2CrystalCal+;
 #pragma link C++ class R3BCalifaMapped2CrystalCalPar+;
 #pragma link C++ class R3BCalifaCrystalCal2TotCalPar+;
-#pragma link C++ class R3BCalifaCrystalCal2Hit+;
+#pragma link C++ class R3BCalifaCrystalCal2Cluster+;
 #pragma link C++ class R3BCalifaCrystalCalDataAnalysis+;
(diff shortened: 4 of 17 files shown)


gabrigarjim:
Hi,  @klenze, maybe  we can use this as an opportunity to improve our data scheme. I was thinking about keeping the hit data, but as a final level. It would be something like

`MappedData -> CrystalCalData -> ClusterData -> HitData `

In this new data level we could store randomized angle, nominal angle, doppler corrected energy, cluster energy and all further reconstruction from ClusterData (maybe also px,py,pz?). This can be achieved by creating another task (i.e CalifaCluster2Hit). I can work on that, as I am already working in the clustering algorithm. What do you think?

hapol:
> The fixed "there are exactly two steps in data processing for any detector" is just not true for many detectors.

This was never fixed. See the [documentation](https://github.com/R3BRootGroup/R3BRoot/wiki/Naming-scheme). Mapped, Cal and Hit level are requested with some characteristics. Many other levels could be added as it is the case in many detectors.

klenze:
@gabrigarjim: [see here](https://github.com/R3BRootGroup/R3BRoot/wiki/CALIFA-data-processing-concept) for my plan. 

@hapol: I see. I still think that Hit is not a good name. If NeuLAND saw four bars firing, I don't think it would be wrong to say it saw four hits. After clustering, it may turn out that it actually saw two Hits, which is kind of confusing. R3BFooPhysicsData might be better.

gabrigarjim:
@klenze , should I comment on that here or directly in the wiki page?

klenze:
I think unlike MediaWiki the github wiki solution is not very good for discussions, so I would prefer to discuss this here or in issues. I have added a disclaimer to the wiki page that this is just my ideas.

gabrigarjim:
Maybe we can open an issue with the "Proposal" label and put there your ideas, so we can discuss.

hapol:
Physical units (cm, ns, MeV) was chosen to reduce the significant ciphers, centering around the size of the relevant units of the experiment. Our experiment delivers distances around cm (from hundreds of microns to meters), times around ns and energies around the MeV.

> (Personally, I would default to the same units as Geant4 does, which happily is MeV. Also, who picked cm as a unit? Are we using cgs now? Geant's unit of mm seems much more reasonable.)

Geant4 unit for energy is GeV (you should know, remember your complains about how slow CALIFA simulation was, when you were sending gammas of several GeV instead MeV some time ago). 
Regarding who picked up cm as a unit, I would say us. Or R3B. It was proposed and accepted long time ago.

jose-luis-rs:
I think that we are only using cm for the "DataPoint" classes used in simulation, for the rest of data levels (and for all the detectors) we are using mm.

jose-luis-rs:
I would like to merge this PR, do you agree?

ryotani:
I have a question. Why do you really need to have R3BCalifaCrystalCalData and R3BCalifaClusterData independent? They are basically just identical. The only difference is you add theta and phi in the latter and CrystalId becomes Cluster Id.
How about calling the new proposed class to be R3BCalifaClusterCalData and making a common data class for R3BCalifaCalData? I think Califa analysis can still be considered with three steps Mapped->Cal'ed->Hit levels.

ryotani:
In the case of plastic detectors, normally it goes as MappedData -> TcalData -> CalData -> HitData. If you say this is "four-step", yes. But it still follows the basic scheme of R3BRoot.

hapol:
@ryotani R3BCalifaClusterData (or present R3BCalifaHitData till this PR is applied) corresponds to a data level where several crystals are add-back to obtain information from a particle depositing energy in more than a single crystal. There is an algorithm applied searching for neighbour crystals and adding energy, with multiple options. 

In a new implementation (not yet public) of R3BCalifaHitData it will contain the list of the crystals contributing to the cluster additionally to the number of crystals contributing, a cluster id, and the theta and phi coordinates determined by an algorithm. By now, the algorithm returns the coordinates theta,phi of the crystal with the maximum deposited energy (either the center of the crystal or randomly on the crystal angular coverage), as it should be the best approach for gammas and it is not a bad approach for protons, but it could/should be improved.

ryotani:
@hapol To me, the Clustering algorithm and your proposing "new HitData" are still in the scope of the Cal level. Although the documentation defines "HIT-level: Physical quantities in cm, ns, MeV", it doesn't explain the difference between the Cal level, as it is also the data level with physical units as MeV.
As you know much better than me, CALIFA is primarily for the detection of charged particles and gamma rays. I believe the Hit level should contain more physics-related information such as the opening angle of two protons, Doppler-corrected gamma-ray energy, or QPID. As the currently proposed ClusteringData (and your proposing new "HitData") doesn't provide any physical information as I listed above. I don't think the values after just adding the energies of some crystals are worth to be categorised as Hit-level. 
I am not against the idea of having different data classes for the CrystalCalData and Cluster"Cal"Data, but what I am proposing here is any data levels below Hit-level and above Mapped-level can be called Cal-level. So, why not calling it ClusterCalData (or CalData)? Then I feel it is more consistent with the naming schemes of other detectors.

hapol:
@ryotani I understand your point now, thank you. It should be considered in the present proposal (and action) of modification of the CALIFA data levels.

jose-luis-rs:
For me the scheme could be R3BCalifaCrystalMappedData -> R3BCalifaCrystalCalData -> R3BCalifaClusterData -> R3BCalifaHitData -> R3BCalifaOthersData

The info for each level should be:
- MappedData: Raw data from ucesb
- CrystalCalData: Calibrated data in MeV, ns, etc per crystal 
- ClusterData: Information related to the variables that characterize the cluster, like the crystals contained in the cluster, size, energy, theta, phi, and QPID. In this case I don't see the need of adding "Cal" because we are not calibrating any variable, just obtaining the parameters that define the cluster properties.
- HitData: This should contain "physics-related information" for tracking, Doppler corrections, etc.
- OtherData like R3BCalifaTrackingData: Reconstruction of the opening angle for protons, opening angle for protons-other particles (n,d,t,..), etc

ryotani:
@jose-luis-rs I don't understand QPID have to be included in ClusterCal class. If you think this is imporant why not including in the CrystalCal class? When analysing data from such scintillator based spectrometers, people have to start looking at reconstruction without clustering (add-back) to valiadte all the method, such as Doppler correction of gamma rays, or proton reconstructions first. If everything is consistent with simulated spectra, then we start looking at the clustering/add-back to gain the statistics and improve the peak-to-total. Taking into this into account, I think all the following reconstructions should be consistent with or without clustering algorithm. Of course people can set add-back length to be zero to disable the clustering algorithm, I would say it is not a good design of the code for the people actually want to use it. That is why I have been saying the DataClass for CrystalCal and ClusterCal should be the same as the Clustering Process should be just an additional process, in other words, a process which can be skipped if someone thinks not necessary. The following process in Hit class, should be able to access the result from CrystalCal or ClusterCal seamlessly, to make further reconstructions. And I think here is the place for QPID and other reconstruction should take place.

jose-luis-rs:
Sounds good @ryotani, I have added the QPID during the cluster algorithm because it will be the last step where we have the info per crystal (QPID is already available at mapped and cal levels). After forming the cluster I am not sure if we will be able to do a PID. If yes, we could add the QPID to the hitData. Did you try something of this with the s467 data?

klenze:
@ryotani: I see where you are coming from with your observation that all fields contained in CalData are also present in ClusterData. I guess making ClusterData inherit from CalData would be one way to reflect that fact. 

If we decide to put the QPID results on a new level, then we can just require CalData for the input of the QPID task, so the user is free to use either CalData or ClusterData. 

IMO, QPID should be most useful when performed at cluster level: A proton which goes through one crystal and then is stopped in a different crystal will hopefully be on the stopped proton branch after clustering, while a proton going through two crystals should end up on the gamma branch. 

(We can also use QPID on Cal level to improve position resolution for protons: if we have two high energy deposits in a cluster, and one has QPID saying "stopped proton", it stands to reason that the other crystal must have seen the proton first, so its position should be used to guess the position where the crystal entered califa, even if it saw less energy than the crystal in which the proton was stopped. Of course, I hope our target tracking will mean that we won't have to resort to this.)

@jose-luis-rs: If you were asking me, I am fine with merging this PR :-)

ryotani:
@jose-luis-rs I have only tried to plot Ns and Nf, so basically I don't know about QPID. I have no idea what kind of algorithms to identify the particles such as proton or deuteron in the CALIFA code. Is it something TCutG parameter?

@klenze Making Cluster(Cal)Data inheriting from CrystalCalData sounds reasonable. Yeah, I was thinking the same as you. I think there is still room for checking QPID at the CrystalCal level as they would be able to tag punch-through or stop.

jose-luis-rs:
This was included with the PR #748. So you could close this PR.

Issue #724, opened 2022-12-14
CI: clang-format todo


klenze:
My proposals for clang-format:
* Move it to a different top level task which gets executed before compilation
* Update .clang-format so that the behavior on lxir136 is compatible with the behavior on the CI machine. 
* Apply clang-format to all files
   * using a git username "whitespace" or something so it is clear in git blame (without -w) 
   * This needs sanity checking of the output, there might be some sections where some user carefully hand-formatted the code and clang-format should be disabled.

ryotani:
Hi, I am not sure this is the right place to add the discussion, but I realised the travis-clang-format-check.sh has been deleted in the recent updates. How can I check the clang format before making PR? Also if we can still have an autofix feature as travis-clang had, that would be great...

klenze:
I moved most of the shell scripts to util/.
util/clang-format-check.sh should still support the autofix option.

Cheers,
   Philipp

jose-luis-rs:
From R3BRoot you can run it as:
. util/clang-format-check.sh

For sofia:
. clangformat_all.sh

I have to generalize it for all the repos, but I am busy with other projects, sorry for that!

ryotani:
@klenze @jose-luis-rs Thanks. It worked!

Issue #674, opened 2022-10-15
Event header, ownership and kicking C-style casting out of R3BRoot


YanzhaoW:
Hello

## Cause:
I stumbled upon a segmentation fault recently during an online analysis using `FairRunOnline`. For most of the time, the segmentation fault could be caused by an unallocated memory being read due to an overflowing index or getting deleted (e.g. a pointer gets deleted twice). The reason why this happened at this time is because the R3B event header was not set in `FairRunOnline` before its initialisation. (Setting the event header at the beginning is certainly not obvious and even less obvious is the idea that the segmentation fault could be caused by this. )

## Explanation:
If the event header is not set manually by the user, `FairRunOnline` would set it automatically to the type `FairEventHeader` during its initialisation:
```cpp
    if (nullptr == fEvtHeader) {
        fEvtHeader = new FairEventHeader();
    }
```
Later it gets registered by `FairRootManager` with:
```cpp
    fRootManager->Register("EventHeader.", "Event", fEvtHeader, (nullptr != fRootManager->GetSink()));
```
In the end, the event header gets deleted in the deconstructor of `FairRun`. Therefore, any other class using the event header, which is obtained through
```cpp
    fHeader = frm->GetObject("EventHeader.");
```
does not own this object and should not delete it. But it gets deleted in the deconstructor of `R3BUnpackReader`:
https://github.com/R3BRootGroup/R3BRoot/blob/628973f43039266af0f833063a943507efeebc61/r3bsource/base/R3BUnpackReader.cxx#L39-L42
In C++, if one pointer gets deleted multiple times, an undefined behaviour would be caused, which could be program crashing, segmentation fault or nothing.

The deletion of not-owned object is just one of the causes. Another cause is the usage of vicious C-style casting, which is omnipresent in R3BRoot. Let me explain it with the following code from `R3BUcesbSource.cxx`:
https://github.com/R3BRootGroup/R3BRoot/blob/4e2decfab13d80ee412c6dfeabfcc19c4efd4976/r3bsource/base/R3BUcesbSource.cxx#L118-L129
[...]

PR #725, opened 2022-12-14
Fixed some instances of C-style casts by replacing them with dynamic_cast

klenze:
Makes some progress to fixing some parts of #674. 

Also, fixed a casting bug which became visible in R3BFieldPar.

Also, I hereby pre-commit to not working with any version of R3BRoot which contains any C-style casts in C++ which were merged after 2022-12-14. 

This code will fail in CI because of clang-format. I will rebase it when #724 is resolved. 

@YanzhaoW could you look over my changes? My regex was not clever enough to know operator precedence, so in theory there could be some instances where ``(T*)A OP B`` was converted to ``dynamic_cast<T*>(A OP B)`` even though the correct replacement would have been ``dynamic_cast<T*>(A) OP B``.  (Of course, this assumes that both cases would compile.) I looked over the replacements, and did not find such cases, but it might help if you double-checked.

### DIFF
--- a/actar/R3BActar.cxx
+++ b/actar/R3BActar.cxx
@@ -201,7 +201,7 @@ Bool_t R3BActar::ProcessHits(FairVolume* vol)
                fELoss);
 
         // Increment number of PspPoints for this track
-        R3BStack* stack = (R3BStack*)gMC->GetStack();
+        R3BStack* stack = dynamic_cast<R3BStack*>(gMC->GetStack());
         stack->AddPoint(kACTAR);
 
         ResetParameters();
@@ -278,7 +278,7 @@ void R3BActar::CopyClones(TClonesArray* cl1, TClonesArray* cl2, Int_t offset)
     R3BActarPoint* oldpoint = NULL;
     for (Int_t i = 0; i < nEntries; i++)
     {
-        oldpoint = (R3BActarPoint*)cl1->At(i);
+        oldpoint = dynamic_cast<R3BActarPoint*>(cl1->At(i));
         Int_t index = oldpoint->GetTrackID() + offset;
         oldpoint->SetTrackID(index);
         new (clref[fPosIndex]) R3BActarPoint(*oldpoint);
--- a/alpide/calibration/R3BAlpideCal2Hit.cxx
+++ b/alpide/calibration/R3BAlpideCal2Hit.cxx
@@ -85,11 +85,11 @@ void R3BAlpideCal2Hit::SetParContainers()
     FairRuntimeDb* rtdb = FairRuntimeDb::instance();
     R3BLOG_IF(fatal, !rtdb, "FairRuntimeDb not found");
 
-    fMap_Par = (R3BAlpideMappingPar*)rtdb->getContainer("alpideMappingPar");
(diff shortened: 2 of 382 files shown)


YanzhaoW:
(complaining begins)
Well, I don't think changing all C-style casting to dynamic cast using regex is a good idea. Dynamic_cast has run-time checks and is considered not fast compared to static_cast.  And afaik C-style casting first performs static_cast, then dynamic_cast and finally interpret_cast. Which means changing all C-style castings to dynamic_cast will slow our software down, especially when you changing it inside Exec() function.

Yeah, changing C-style casting to dynamic_cast at least can make sure we have less segmentation fault. The "ideal" case is to change it accordingly, which needs tons of manpower and time. And even if we have that manpower and time, why not get rid of those casting once for all? **Frequent usages of casting only means the code is not "well written" and poorly designed.**   
(complaining ends here)

Nevertheless, let's have a try. I will look into the errors. What's your regex? Is it possible that we don't do the changing in the body of some specific functions like : Exec(), Finish() ...

klenze:
I think dynamic_cast is the correct solution for us in most cases. 
With regard to looping over elements of TCAs in Exec, I would like to state that the placement new used with TClonesArrays has no type checking. (Of course, this also means that we get overlapping objects if we instantiate a larger object than the TCA expects. Perhaps valgrind/libasan could catch that.) 

I think C-style casts have no place in C++, and a dynamic_cast is a reasonably safe replacement for our uses. The overhead is the price we pay for bad design. 

I would be willing to discuss using static_casts under the following conditions:
0. The code in question is otherwise optimized for performance. 
1. There is a convincing reason why a type-safe construct can not be used. 
2. Guarantees are in place which will prevent type errors.
3. A performance analysis shows that the speed impact is significant.

(For example, I could possibly be persuaded that the my TCA wrapper fulfills conditions 1. (backward compatibility) and 2. (template type is checked against the TCA type during construction))

In my experience, the code I read is often clearly not optimized for performance: Tasks which could be accomplished in O(n*log(n)) (such as "finding pairs of entries") are implemented naively in O(n*n) instead. 

Also, I think that different levels of danger need different levels of care: ASM > C > (clean) C++ > Java/Python. 
If one writes in C on needs to be much more careful than in Python. 
Given our level of care, I think it would have been better to use a memory safe language such as Java or Python in the first place. static_cast basically means moving towards the bare metal end of the spectrum, and I don't think our contributors are ready for that. 

Getting rid of the casting is not that easy. People expect to be able to serialize objects to TTrees and use them in their ROOT macros. If a TTree can contain STL types, we could migrate to them (and use map<std::string, boost::any> or something for lookup by name), but this will take some work for each Task. The dynamic_casts are a quick fix. 

I think the (extended) RE I used was something like ``\( *(R3B[A-Za-z]* *[*] *)\)(.*);`` and a replacement like ``dynamic_cast<\1>(\2);`` (from memory, HISTFILESIZE was to small. Plus some other cases for if clauses). 

Surprisingly, this did not yield too many false positives. (Doing the same with the Fair prefix caught method declarations like ``Foo(FairBar*);``, so I just added the one cast case I found manually.)

klenze:
redone with clang-tidy in #759.

PR #727, opened 2022-12-15
Checking travis

jose-luis-rs:
Don't merge!


jose-luis-rs:
Here I am doing other tests ... I think that using -Werror we will find more than 500 warnings

PR #728, opened 2022-12-15
Rewriting Neuland digitisation simulation module (Part 2/2)

YanzhaoW:
This is the final major change to the neuland digitisation module, which is mainly about improving the code architecture. Here is a list of changings:

1. Change class `DigitizationEngine` into a class template:

    With this, its implementation to other detectors are much easier and there is no need to create additional child class. The default digitisation engine in `R3BNeulandDigitizer` is still neuland bar with Tacquila channel:
   ```cpp
   using DefaultEngine = DigitizingEngine<NeulandPaddle, TacquilaChannel>;
   
   R3BNeulandDigitizer::R3BNeulandDigitizer(TString input, TString output)
    : R3BNeulandDigitizer(std::make_unique<DefaultEngine>(), std::move(input), std::move(output))
   {
   }
   ```
   Everything in the Root MACRO should still works.
   If a digitizer with Tamex Channel is needed, simply do: 
   ```cpp
   using NeulandPaddle = Digitizing::Neuland::NeulandPaddle;
   using TamexChannel = Digitizing::Neuland::Tamex::Channel;
   using Digitizing::DigitizingEngine;
   using digitizingEngTamex = DigitizingEngine<NeulandPaddle, TamexChannel>;

   auto digiEng = std::make_unique<digitizingEngTamex>();
   run->AddTask(new R3BNeulandDigitizer(std::move(digiEng)));
   ```

2. Move class `Digitizing::Channel` and `Digitizing::Paddle` into separate files.
3. A new class named `DigitizingPaddleNeuland` is created and many functionalities of the old `Digitizing::Paddle` are moved into this new class.
4. Create a new exectuble neulandAna, which runs this new digitization module with neuland paddle and Tamex channel.

## problems to be solved in the future
1. ROOT macro doesn't support class template and variadic function template. What I could do, according to ROOT manual, is to specify every possible instantiation of the class template in the LinkDef.
2. DigitizingEngine is now more versatile but at the cost of user friendliness. Suppose I have my own class for paddle and channel: 
   ```cpp
   class MyPaddle: public Paddle{
   public:
      MyPaddle(double v);
    };
   class MyChannel : public Channel{
   public:
      MyChannel(double v);
    };
   ```
   To use with the DigitizingEngine, it's a bit complicated:
   ```cpp
   using DigiEngClass = DigitizingEngine<MyPaddle, MyChannel, std::tuple<double>, std::tuple<double>>;
   auto DigiEng = std::make_unique<DigitClass>({v1}, {v2});
   run->AddTask(new R3BNeulandDigitizer(std::move(DigiEng)))
   ```
It seems mixing implicit and explicit instantiation of class template is not possible. I will figure out how to mitigate these complication.

## Todo
1. Still writing the documentation and hopefully I will add it before the beginning of the next year.
2. Presentation about this will be given in the Analysis WG meeting in the next year.

Any comments or suggestions are welcomed. ;D

Edit: the first commit is when DigitizingEngine is an abstract class rather than a class template.


YanzhaoW:
@jose-luis-rs Is it possible that we upgrade our CI test to C++17?

jose-luis-rs:
It depends on the FairSoft and FairRoot versions installed somewhere..., so this upgrade will require some inputs by @kresan

klenze:
> ROOT macro doesn't support class template and variadic function template. What I could do, according to ROOT manual, is to specify every possible instantiation of the class template in the LinkDef.

I think the solution here is not to use ROOT macros. :-)
If 99% of the C++ users compile their code, we should consider the possibility that they are correct in doing so. 
(In practice, this would mean having universal main functions which can parse some steering file to know what to instantiate. Of course, these compile units would also need to know what instances of the template classes they need to generate.)

> (you wrote)
```
using DigiEngClass = DigitizingEngine<MyPaddle, MyChannel, std::tuple<double>, std::tuple<double>>;
auto DigiEng = std::make_unique<DigitClass>({v1}, {v2});
run->AddTask(new R3BNeulandDigitizer(std::move(DigiEng)))
```

I think it would be useful to have a templated helper function which does this. Something like, I dunno,
```
template<class Paddle, class Channel, class... PadArgs, class ... ChanArgs>
decltype(auto) makeDigiEng(std::tuple<PadArgs...> pa, std::tuple<ChanArgs...> ca)
{
   return std::move(std::make_unique<DigitizingEngine<Paddle, Channel, 
                   std::tuple<PadArgs...>, std::tuple<ChanArgs...> > (pa, ca) >);
}
```
Of course, it is totally possible that this will not work. Or not pass by rvalue reference as intended.

Or you could use factory closures (do such things exist?) to use pass the state on:
```
template<class T,  class... Args>
auto makeFactory(Args... pargs)
{
  return [pargs...](auto... args){ return new T(args..., pargs...);}; // or a smart pointer if you like. 
}

template<typename PaddleFactory, typename ChannelFactory>
class DigitizingEngine
{
   using Paddle=decltype(PaddleFactory(0));
   using Channel=decltype(Channel(0));
   PaddleFactory BuildPaddle;
   ChannelFactory BuildChannel;
   DigitizingEngine(PaddleFactory pf, ChannelFactory cf) // or pass ownership by unique_ptr or whatever
   : BuildPaddle(pf)
   , BuildChannel(cf)
   {
   }
};
// in that case, makeDigiEngine is left to the reader. 
run->AddTask(new R3BNeulandDigitizer(makeDigiEngine( createFactory<PaddleClass>(), // no additional state for paddle here
                                                     createFactory<Channel>(42, "more state")  // but some to pass to the channel constructor

                                    )))
```
(See [test.txt](https://github.com/R3BRootGroup/R3BRoot/files/10249947/test.txt) for some of that. )

Of course, some people might claim that having functions which return functions which return object pointers is a bit complicated, while others will complain that the new operator in makeFactory is too rigid and and should have the template parameters ``<class T,  auto op=newop<T>, class... Args >`` where ``newop<T>`` will be replaced with ``[](auto...args){return new T(args...);}`` as soon as g++ allows variadic lambdas as default arguments for template parameters.  :-)

I will try to look through your code a bit after the weekend. Yesterday I had to look up what mutable does exactly.

YanzhaoW:
@klenze good idea. See the new commit.

Though I made few changes from what you suggested:
```cpp
    // factory classes for paddle and channel:
    template <typename ChannelClass,
              typename = typename std::enable_if<std::is_base_of<Channel, ChannelClass>::value>::type>
    class UseChannel
    {
      public:
        template <typename... Args>
        explicit UseChannel(Args&&... args)
            : BuildChannel([&](Channel::SideOfChannel side) {
                return std::make_unique<ChannelClass>(side, std::forward<Args>(args)...);
            })
        {
        }
        std::function<std::unique_ptr<ChannelClass>(Channel::SideOfChannel)> BuildChannel;
    };

    template <typename PaddleClass,
              typename = typename std::enable_if<std::is_base_of<Paddle, PaddleClass>::value>::type>
    class UsePaddle
    {
      public:
        template <typename... Args>
        explicit UsePaddle(Args&&... args)
            : BuildPaddle(
                  [&](size_t paddleID) { return std::make_unique<PaddleClass>(paddleID, std::forward<Args>(args)...); })
        {
        }
        std::function<std::unique_ptr<PaddleClass>(size_t)> BuildPaddle;
    };

    template <typename PaddleClass, typename ChannelClass>
    class DigitizingEngine : public DigitizingEngineInterface
    {
      private:
        UsePaddle<PaddleClass> paddleClass;
        UseChannel<ChannelClass> channelClass;
        std::map<size_t, std::unique_ptr<Paddle>> paddles;

      public:
        DigitizingEngine(const UsePaddle<PaddleClass>& p_paddleClass, const UseChannel<ChannelClass>& p_channelClass)
            // explicit DigitizingEngine(const PaddleArgsTuple& p_args = PaddleArgsTuple{},
            //                           const ChannelArgsTuple& c_args = ChannelArgsTuple{})
            : paddleClass{ p_paddleClass }
            , channelClass{ p_channelClass }
            , DigitizingEngineInterface()
        {
        }
    // ...
    };


```
This makes creation of digitizingEngine much nicer and clearer:
```cpp
using NeulandPaddle = Digitizing::Neuland::NeulandPaddle;
using TamexChannel = Digitizing::Neuland::Tamex::Channel;
using Digitizing::UseChannel;
using Digitizing::UsePaddle;

auto tamexEngine = Digitizing::CreateEngine(UsePaddle<NeulandPaddle>(), UseChannel<TamexChannel>());
run->AddTask(new R3BNeulandDigitizer(std::move(tamexEngine)));
```

klenze:
Looks good. 

Two minor nitpicks:
* std::function has some overhead over raw lambdas. [this measurement](https://artificial-mind.net/blog/2019/09/07/std-function-performance) claims a difference of 5 cycles between funD and funH. If this is a price you are willing to pay for avoiding having template parameters to capture the types of some lambda I am totally ok with that. 

* Personally, I like to use std::enable_if only to exclude alternatives in SFINAE contexts. Here, enable_if will probably not prevent me from explicitly instantiating ``UseChannel<void, void>``. static_assert will work until we get concepts.

YanzhaoW:
> Looks good. 
> 
> 
> 
> Two minor nitpicks:
> 
> * std::function has some overhead over raw lambdas. [this measurement](https://artificial-mind.net/blog/2019/09/07/std-function-performance) claims a difference of 5 cycles between funD and funH. If this is a price you are willing to pay for avoiding having template parameters to capture the types of some lambda I am totally ok with that. 
> 
> 
> 
> * Personally, I like to use std::enable_if only to exclude alternatives in SFINAE contexts. Here, enable_if will probably not prevent me from explicitly instantiating ``UseChannel<void, void>``. static_assert will work until we get concepts. 

Thanks for the tips.

• I will stay with std::function since function pointer is too ugly to read. 
• God forbids ever passing void type anywhere in C++

jose-luis-rs:
You can check your tests using the command:
>> make test
from the folder build

YanzhaoW:
@kresan 

C++17 CI container seems to have following problems:
1. does not have gtest installed, which leads to all Gtest untested.
3. can't run executable. This is probably due to the lack correct version of libstdc++ in LD_LIBRARY_PATH.

Could you fix that?

kresan:
@YanzhaoW 

1. I can have a look at how to provide gtest.

2. Which executable? How do you run it? What is the error message? Be more precise.
I do not think that one needs to put libstdc++ from system path into LD_LIBRARY_PATH to be able to run an executable. Something is fishy here.

YanzhaoW:
> @YanzhaoW
> 
> 1. I can have a look at how to provide gtest.
> 2. Which executable? How do you run it? What is the error message? Be more precise.
>    I do not think that one needs to put libstdc++ from system path into LD_LIBRARY_PATH to be able to run an executable. Something is fishy here.

Thanks for the reply

The executable is `build/bin/neulandAna`. To run it, just simply execute it with 
```shell
./neulandAna
```
The error message is simply just test failing. I can't get more messages from the C++17 CI.

I had the same problem of running the executable in lxir136. And later I found out that the library path /usr/lib/ contains a very old libstdc++. After I add the library path of new libstdc++ to the LD_LIBRARY_PATH, it works. Which version of libstdc++ do we use in the CI container?

kresan:
In the CI container we have Debian 10 with standard gcc 8.3 installed, so libstdc++ 3.4.25.

YanzhaoW:
> In the CI container we have Debian 10 with standard gcc 8.3 installed, so libstdc++ 3.4.25.

listdc++ has version 3.4. Is it the correct version for gcc8.3?  In this [list](https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html), the version of libstdc++ is 6.0 for gcc 8.1. The executable works with libstdc++ 6.0.28 with gcc 10.

Second, is there a way I can get the verbose output of ctest in C++17 CI? The link in the standard output seems to be broken.

jose-luis-rs:
Please, rebase dev

neuland/digitizing/DigitizingChannel.cxx
```diff
@@ -0,0 +1,54 @@
...
+        return fSignals.getRef();
+    }
+
+    auto Channel::HasFired() -> bool { return (!GetSignals().empty()); }
+
+    auto Channel::GetTrigTime() -> double
+    {
+        if (!fTrigTime.valid())
+        {
+            const auto& signals = GetSignals();
```
jose-luis-rs: Sometimes is dificult to understand the code because one cannot identify the objects created with "auto", what kind of object is GetSignals()?
const Foo& signals = GetSignals();
will be better.

neuland/digitizing/DigitizingChannel.cxx
```diff
@@ -0,0 +1,54 @@
...
+        if (!fSignals.valid())
+        {
+            fSignals.set(ConstructSignals());
+        }
+        return fSignals.getRef();
+    }
+
+    auto Channel::HasFired() -> bool { return (!GetSignals().empty()); }
+
+    auto Channel::GetTrigTime() -> double
```
jose-luis-rs: double Channel::GetTrigTime() is shorter, 
Why have we to write 6 extra characters for a simple definition?

neuland/digitizing/DigitizingTacQuila.cxx
```diff
@@ -12,175 +12,168 @@
...
+            // Maximum pulse height or just sum over all light = charge?
+            Double_t light = 0.;
+            for (auto hit_it = cachedFirstHitOverThresh.get(); hit_it != fPMTHits.end(); hit_it++)
             {
-                // @mheil: Should that be calculated using an exponential with the
-                // prior accumulated light / exponential decay or just all the light that arrives?
-                // Maximum pulse height or just sum over all light = charge?
-                Double_t light = 0.;
-                for (auto hit_it = cachedFirstHitOverThresh.get(); hit_it != fPMTHits.end(); hit_it++)
+                const auto hit = *hit_it;
```
jose-luis-rs: Again, auto makes difficult the understanding.

neuland/executables/neulandAna.cxx
```diff
@@ -0,0 +1,129 @@
...
+    TStopwatch timer;
+    timer.Start();
+
+    auto channelName = std::string{ "tacquila" };
+    auto paddleName = std::string{ "neuland" };
+    auto paraFileName = std::string{ "para.root" };
+    auto simuFileName = std::string{ "simu.root" };
+    auto digiFileName = std::string{ "digi.root" };
+    auto logLevel = std::string{ "error" };
+    auto eventNum = 0;
```
jose-luis-rs: what is eventNum? 
int, double, float?

neuland/executables/neulandAna.cxx
```diff
@@ -0,0 +1,129 @@
...
+    {
+        logLevel = varMap["logLevel"].as<std::string>();
+    }
+    if (varMap.count("eventN") != 0U)
+    {
+        eventNum = varMap["eventN"].as<int>();
+    }
+    FairLogger::GetLogger()->SetLogScreenLevel(logLevel.c_str());
+
+    auto run = new FairRunAna{};
```
jose-luis-rs: pointer?

neuland/executables/neulandSim.cxx
```diff
@@ -0,0 +1,143 @@
...
+#include <boost/program_options.hpp>
+#include <iostream>
+#include <string>
+
+namespace po = boost::program_options;
+
+int main(int argc, const char** argv)
+{
+    TStopwatch timer;
+    auto eventNum = 100000;
```
jose-luis-rs: Here I don't know if these variables are int, float, double ....

neuland/digitizing/DigitizingChannel.cxx
YanzhaoW: One can easily identify the type with the help from some static analysis tools.

jose-luis-rs: Could you provide some examples?

YanzhaoW:
@jose-luis-rs  `auto` has already been preferred in C++ community since C++11 (12 years ago). The basic principle is using `auto` as often as possible. I have talked about this in the last R3B analysis meeting and I will talk about it again if necessary in next ECR meeting. So I hope anyone who didn't go to last R3B analysis meeting come to ECR meeting next week.

The biggest advantage using `auto` is making implicit conversion almost impossible. And it's also to write and makes code easy to change. There is no disadvantage of using `auto`. Your concern about readability can be solved with IDE or external static analysis tools. I mean those who are a little bit serious about coding should at least have one or more such kind of tools at hand.

For the ECR meeting next week, I will introduce one of the tools: clangd implemented in Neovim. If someone doesn't like vim or neovim, just pick VSCode.

neuland/digitizing/DigitizingTacQuila.cxx
YanzhaoW: I only implemented tamex code. This was written by someone else long time ago. But I will only fix all the warnings relating to tacquila implementation.

neuland/digitizing/DigitizingChannel.cxx
YanzhaoW: consistency 

plus easy to read.

neuland/executables/neulandAna.cxx
YanzhaoW:
```cpp
auto var1 = 3;  // var1 is int
auto var2 = 3.; //var2 is double
auto var3 = 3.f; // var3 is float
auto var4 = 3L; // var4 is long
```
Please check the literal suffix in C++. It's very easy to understand if you know it.

neuland/executables/neulandAna.cxx
YanzhaoW: Yes, I just copied it from root Macro. Clang-tidy also reports warnings for this.

neuland/executables/neulandSim.cxx
YanzhaoW: Same above. It's `int`.
I can do this as well:
```cpp
auto const eventNum = int{100000};
```

jose-luis-rs:
> @jose-luis-rs `auto` has already been preferred in C++ community since C++11 (12 years ago). The basic principle is using `auto` as often as possible. I have talked about this in the last R3B analysis meeting and I will talk about it again if necessary in next ECR meeting. So I hope anyone who didn't go to last R3B analysis meeting come to ECR meeting next week.
> 
> The biggest advantage using `auto` is making implicit conversion almost impossible. And it's also to write and makes code easy to change. There is no disadvantage of using `auto`. Your concern about readability can be solved with IDE or external static analysis tools. I mean those who are a little bit serious about coding should at least have one or more such kind of tools at hand.

Ok, you mean that all our PhD students must use "IDE or external static analysis tools" to understand the code and do the data analysis, right?

> 
> For the ECR meeting next week, I will introduce one of the tools: clangd implemented in Neovim. If someone doesn't like vim or neovim, just pick VSCode.

YanzhaoW:
> Ok, you mean that all our PhD students must use "IDE or external static analysis tools" to understand the code and do the data analysis, right?

I mean even bachelor students use IDE for coding. I'm pretty sure it won't be difficult for PHD students in our collaboration.

@janmayer

janmayer:
YOU CALLED?

> Ok, you mean that all our PhD students must use "IDE or external static analysis tools" to understand the code and do the data analysis, right?

Uhm, yes, of course? What are they opening the code with, Notepad? VS Code exists. Demand the best tools available.

Yanzhao's comment about `auto` is correct.

I haven't looked at the PR at all, but don't overengineer things to much. Especially abstractions and templates and so on are only useful if that functionality is used in many places. I've read the word variadic template somewhere, and got scared. Man, I am glad not to work with C++ anymore. How about y'all rewrite the whole thing in Rust?

YanzhaoW:
> YOU CALLED?
> 
> 
> 
> > Ok, you mean that all our PhD students must use "IDE or external static analysis tools" to understand the code and do the data analysis, right?
> 
> 
> 
> Uhm, yes, of course? What are they opening the code with, Notepad? VS Code exists. Demand the best tools available.
> 
> 
> 
> Yanzhao's comment about `auto` is correct.
> 
> 
> 
> I haven't looked at the PR at all, but don't overengineer things to much. Especially abstractions and templates and so on are only useful if that functionality is used in many places. I've read the word variadic template somewhere, and got scared. Man, I am glad not to work with C++ anymore. How about y'all rewrite the whole thing in Rust?

Very glad you have the similar opinion. But don't forget to add another entry in this (repo)[https://github.com/ansuz/RIIR] :D

jose-luis-rs:
> YOU CALLED?
> 
> > Ok, you mean that all our PhD students must use "IDE or external static analysis tools" to understand the code and do the data analysis, right?
> 
> Uhm, yes, of course? What are they opening the code with, Notepad? VS Code exists. Demand the best tools available.
> 
> Yanzhao's comment about `auto` is correct.
> 
> I haven't looked at the PR at all, but don't overengineer things to much. Especially abstractions and templates and so on are only useful if that functionality is used in many places. I've read the word variadic template somewhere, and got scared. Man, I am glad not to work with C++ anymore. How about y'all rewrite the whole thing in Rust?

@janmayer: I think that @YanzhaoW has had a good idea to discuss this point, let's see the results!

klenze:
> • I will stay with std::function since function pointer is too ugly to read.

To clarify: Function pointers are a syntactical abomination. They have less of a place in C++ user code than malloc. 
I was probably thinking of passing lambda expressions as an auto parameter or auto template parameter. Of course, in your case it would have to be a template parameter, and you may well decide that the syntactical burden is not worth it. Never optimize before measuring and all that. 

Speaking of auto, it seems I can't spend two weeks teaching without you having a heated discussion ... without me!

Time for some more hot takes:

* I think we should strive to make the program **flow** readable from the code itself, without resorting to compilers. If the type of x is used to determine which version of f gets called, then the type of x should be obvious from reading the code. 
* I think in contexts where the exact type is important, I prefer
```uint64_t x{};      // (0)```
to 
```auto x=0UL;        // (1) bad, assumes sizeof(unsigned long)==8 ```
or the longer
```auto x=uint64_t{}; // (2) longish but acceptable ```
(The reason why I dislike the latter is that the syntactical building block of ``<concrete type> <variable name>`` will stay with C++ for a long time, in function/method declarations at least, so every programmer will have to be able to parse them, so one might as well use it elsewhere to shorten the code.) My proposal would be to permit both (0) and (2).
* OTOH, I strongly prefer to auto-deduce the type if the type is long and boring, such as for a lambda expression. I prefer auto if the RHS of the initialization is a new statement (and certainly if the RHS is make_unique or something). 
* For many cases in between, I can't really bring myself to care either way. If you declare a loop counter as ``int i{}``, I parse that as "I don't care about the exact bit size of i (else I would have used stdint)". In that case, ``auto i=0;`` (or ``auto i=0U``) seems about the same. 
* If you do need to declare a raw pointer initialized to nullptr, I guess the auto syntax would become:
```auto* ptr=static_cast<Foo*>(nullptr); ```
I much prefer 
```Foo* ptr{};```
(Some may argue that discouraged semantics deserve painful syntax, though. Perhaps that is the rationale behind function pointers as well.)

How about a compromise? Let's split the land along the size of the variable. 
For short variables (equal or less than 32 bits), we prefer short declarations:
```char c{}; uint16_t s{};```
while for long variables (larger sizes), we prefer long declarations:
```auto z=uint64_t{}; auto* ptr=static_cast<Foo*>(nullptr); // on amd64```
(I jest, of course.)

**Comment by YanzhaoW** [comment]
@klenze Ok, I agree using `auto` for simple initialisations such as integers is about the personal taste and it doesn't worth arguing about it.
```cpp
int i = 1;
auto i = 1;
auto i = int{1};
```
Pick whichever you want. (Now I prefer the third one since it's safer than the second.)

But for something like:
```cpp
auto signal = GetSignal();
```
`auto` is clearly better than writing the type out. First it's short. Second, which is more important, it prevents any possibility of implicit conversion. For this reason, I think everyone should use this. And of course, it's also easy for me if I want to change the return type of GetSignal().

YanzhaoW:
@klenze Ok, I agree using `auto` for simple initialisations such as integers is about the personal taste and it doesn't worth arguing about it.
```cpp
int i = 1;
auto i = 1;
auto i = int{1};
```
Pick whichever you want. (Now I prefer the third one since it's safer than the second.)

But for something like:
```cpp
auto signal = GetSignal();
```
`auto` is clearly better than writing the type out. First it's short. Second, which is more important, it prevents any possibility of implicit conversion. For this reason, I think everyone should use this. And of course, it's also easy for me if I want to change the return type of GetSignal().

YanzhaoW:
@klenze Ok, I agree using `auto` for simple initialisations such as integers is about the personal taste and it doesn't worth arguing about it.
```cpp
int i = 1;
auto i = 1;
auto i = int{1};
```
Pick whichever you want. (Now I prefer the third one since it's safer than the second.)

But for something like:
```cpp
auto signal = GetSignal();
```
`auto` is clearly better than writing the type out. First it's short. Second, which is more important, it prevents any possibility of implicit conversion. For this reason, I think everyone should use this. And of course, it's also easy for me if I want to change the return type of GetSignal().

YanzhaoW:
Here is a video for everyone about `auto` in a talk given by Herb Sutter in Cppcon 14. He begins to talk about `auto` starting from [28:22](https://youtu.be/xnqTKD8uD64?t=1702).

With this, I hope we could close our discussion about `auto`.

YanzhaoW:
@jose-luis-rs This PR is ready to be merged.

neuland/digitizing/CMakeLists.txt
```diff
@@ -11,25 +11,35 @@
...
+    DigitizingTacQuila.cxx
+    DigitizingTamex.cxx
+    DigitizingPaddleNeuland.cxx
+    R3BNeulandHitMon.cxx
+    R3BNeulandDigitizer.cxx
+    DigitizingPaddleMock.h
+    DigitizingChannelMock.h)
+
+set(HEADERS
+    DigitizingChannel.h
```
jose-luis-rs: In R3BRoot the name of the classes should start with "R3B", 
why do these classes:
DigitizingChannel.cxx
DigitizingTacQuila.cxx
DigitizingPaddle.cxx
DigitizingTamex.cxx
DigitizingTacQuila.cxx
DigitizingTamex.cxx
DigitizingPaddleNeuland.cxx
break the agreement?

YanzhaoW: Hi,

Ok, sorry I didn't know we have this rule. But is this agreement only for the file name or the class name?

But sure I will change the file name first.

jose-luis-rs: Both, file name and the class name.

YanzhaoW: But does this agreement also apply to the class inside a namespace? Such as 
```cpp
namespace Digitizing::Neuland
{
    class PaddleHit;
}
```
And another question is whether we also have this agreement on the namespace? Because if we only have them on the class name, it would look very funny:
```cpp
Digitizing::Neuland::R3BPaddleHit;
```
Instead I would say, following is much better
```cpp
R3BDigitizing::Neuland::PaddleHit;
```

klenze: We mostly do not use name spaces because ROOT did not do so in ~1995. 

I would argue that it should be 
```R3B::Neuland::Digitizing::PaddleHit```

Some people would probably argue that there should be a correspondence between the namespace path and the file path. 

I would suggest the following:
* For serializable data, keep the current class naming scheme. 
* For tasks, at least provide an alias:
``
using R3BCalifaMapped2CrystalCal = R3B::Califa::Mapped2Cal;
``
(No idea if this will work well with all the dark magic needed to invoke the tasks within "C scripts". It might be safer to also stick to the legacy naming convention.)
* For implementation details which are not important for people running macros, I think there is no reason not to use naming conventions from our current century instead.

Comment (jose-luis-rs) [reply neuland/digitizing/CMakeLists.txt]: For the classes the rule is:
R3B+"Detector name"+"specific names"

For the namespace, please, try to look for the best to ensure that our students can later understand what the code does

Comment (YanzhaoW) [reply neuland/digitizing/CMakeLists.txt]: I would like to discuss a little bit more about `Neuland::Digitizing` and `Digitizing::Neuland`.

If we use `Neuland::Digitizing`, which namespace should we put some common digitizing base classes that can be used in many different detectors? (Like for Neuland and TofD, they both use bar shaped scintillators and similar digitizers.) So normally we put them into a namespace `R3B::Digitizing`. Now if some algorithms from Neuland defined in the `R3B::Neuland::Digitizing` need to use those base classes from `R3B::Digitizing`, they need to go all the way to the top level, like:
```c++
namespace R3B::Neuland::Digitizing
{
    class MyClass : public R3B::Digitizing::Base {};
}
```
But on the other hand, if we use `R3B::Digitizing::Neuland`, that becomes very clean:
```c++
namespace R3B::Digitizing::Neuland
{
    class MyClass : public Base {};
}
```
With this naming style, we could also reduce some naming chaos like `R3B::Neuland::Digitizing`, `R3B::Tofd::digitizing` or `R3B::Califa::Digi`. Instead, they could just be `R3B::Digitizing::Neuland`, `R3B::Digitizing::Tofd` or `R3B::Digitizing::Califa`.

YanzhaoW: @jose-luis-rs Sure, but is `R3B::Digitizing::Neuland::PaddleHit` suitable?

jose-luis-rs: > I would like to discuss a little bit more about `Neuland::Digitizing` and `Digitizing::Neuland`.
> 
> If we use `Neuland::Digitizing`, which namespace should we put some common digitizing base classes that can be used in many different detectors? (Like for Neuland and TofD, they both use bar shaped scintillators and similar digitizers.) So normally we put them into a namespace `R3B::Digitizing`. Now if some algorithms from Neuland defined in the `R3B::Neuland::Digitizing` need to use those base classes from `R3B::Digitizing`, they need to go all the way to the top level, like:
> 
> ```c++
> namespace R3B::Neuland::Digitizing
> {
>     class MyClass : public R3B::Digitizing::Base {};
> }
> ```
> 
> But on the other hand, if we use `R3B::Digitizing::Neuland`, that becomes very clean:
> 
> ```c++
> namespace R3B::Digitizing::Neuland
> {
>     class MyClass : public Base {};
> }
> ```
> 
> With this naming style, we could also reduce some naming chaos like `R3B::Neuland::Digitizing`, `R3B::Tofd::digitizing` or `R3B::Califa::Digi`. Instead, they could just be `R3B::Digitizing::Neuland`, `R3B::Digitizing::Tofd` or `R3B::Digitizing::Califa`.

I think that the digitizers for all these detectors are completely different, what is the common part?

YanzhaoW: @jose-luis-rs There are a lot of common parts, like the two side readouts, scintillation, light attenuation, etc. Even if they are very different, they could derived from a shared base class. For example, the simulations of different detectors come from the same base class `R3BDetector`.

jose-luis-rs: Yes, but in the case of heavy ions the detection efficiency is around 99% and the "two side readouts, scintillation, light attenuation, etc." are redundant, basically because we only want to determine basic things like the geometrical efficiency.

Of course, what you say is somehow important if one wants to have a real detector response, in particular for the lighter ions. It could be interesting to have a general digitizer class able to manage the digitization of all kinds of particles and provide functions and methods for the other digitizers.

YanzhaoW:
@jose-luis-rs Namespaces have now been changed to have "R3B" as the start. If there is no other problem, this PR is ready to be merged.

jose-luis-rs:
> @jose-luis-rs Namespaces have now been changed to have "R3B" as the start. If there is no other problem, this PR is ready to be merged.

I can see a new class R3B::Digitizing::Neuland::TacQuila, are we still using the Tacquila electronics for NeuLAND?

igasparic:
> > @jose-luis-rs Namespaces have now been changed to have "R3B" as the start. If there is no other problem, this PR is ready to be merged.
> 
> I can see a new class R3B::Digitizing::Neuland::TacQuila, are we still using the Tacquila electronics for NeuLAND?

Although we are not using it any more I would keep it because most of the simulations so far have been done with this digitizer.

PR #741, opened 2023-01-03
[Merge Ready] Now Compiling with g++12 -std=c++20, FR 18.8.

klenze:
Added missing includes: FairRootManager and TCA

Fixed (?) a lambda expression in neuland. 

@YanzhaoW: Can you look at my neuland fix if this is acceptable to you?


klenze:
@kresan @jose-luis-rs: Is there some mechanism to put larger files for retrieval on the web? I previously tried pastebin sites, but they refuse to work with a megabyte of compile logs, and clearly my filtering script is not working here.

neuland/simulation/R3BNeulandMCMon.cxx
```diff
@@ -505,14 +505,16 @@ void R3BNeulandMCMon::Exec(Option_t*)
...
+                    tracks.begin(), tracks.end(), TString(), [&](TString&& s, const R3BMCTrack* b) -> TString&& {
                         if (std::find(filteredPdgCodes.begin(), filteredPdgCodes.end(), b->GetPdgCode()) !=
                             filteredPdgCodes.end())
                         {
-                            return s;
+                          return std::move(s);
                         }
-                        return s += "_" + TString::Itoa(b->GetPdgCode(), 10);
+                        return std::move(s += "_" + TString::Itoa(b->GetPdgCode(), 10));
                     });
```
YanzhaoW, line 519: There is a problem indeed. Return by value not by rvalue reference.
So things should be like:
```cpp
std::accumulate(tracks.begin(), tracks.end(), TString{}, 
                      [&](auto && s, auto &b) -> TString
                       {
                           //...
                           return s; // compiler automatically moves (check return value optimisation.)
                       }                 
)
```
If you specify the return value with rvalue, you will get nothing in the end. Check this [test](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAOykrgAyeAyYAHI%2BAEaYxBKBAA6oCoRODB7evgGkyamOAqHhUSyx8VyBdpgO6UIETMQEmT5%2BlZj2BQx1DQRFkTFxCbb1jc3ZlSO9Yf2lgxUAlLaoXsTI7BwA9BsA1AAqAJ6JmNv7y8TbaFjbCHGYpNsk27SoTOjbhtuYqqyJ9AB0Jg0AEFzABmMLIbxXEygtwMHxxPCmUHYQEgszghiQrzQ2EKAjEMLAGGo4FgiFQ44wtwAN2qRGIJLR5KxlO21Kc%2BOImFYTLJwKYXiI2xYTDCEHm2wAtCTtmECOyyf4rMDtmr3kLUNsuUSAGrVdmggAi7OV5jMN1oz3N93N7LMZltDoA7iRaOgbfazNswGBzSZ/EaYSqQar1VtttyCCsGNtovttjSxF5MGj1RrhdyFIaTfj0CAQExkMgfF4DARMBAdUZ9cg/rFgOLFoqgenq8Ba39XBL7nmC%2B2A1ZA7aw2qTABWKxmABsE6NEEFwvMs5n20SAH0woR7mgGPiM6hl3KCJKZSjtQR8yAB2bR%2Bm15uGIR7VZjcfg2n72qozGH1uCB%2Bd4Bka8yAa26p9igywKtS1JejQxD7lmZYECAXqGm4cFZhhcGQa4tBgcy/LgWqEY/sQsbEEm3jsJ%2BaqLlqWbrt6MK5peBZFiWLBlkwFZVgSerVPWmCNgwPYtm2Ak1kJ3bNpBN5DkaI4keyk7LnOC6avaK7Tn%2BT4EDuAj7gxR6EKesryVJxIzkeg50feG7/i%2BrHvqCIZft%2BmDRhRF5XiwqB0hAjn6aBbn2cBoXuRB7FQUKOGwl6Si7m8yG0Kh6GwQlTEsbCuExfhhGBhwiy0Jw468H4HBaKQqCcJhljWNqZxrPaoI8KQqFVcViwANYgOO/h/JIAAc/jTpI/hmMNZiSBoZhcKC%2BicJIFWaLwtUcLwCggBoHVrYscCwDAiBQSwiR0HE5CUGgZ0XfEwBcA6pBYDSSKYLqeCYM6ADyRyVe1NBpXE20QNEa2kNEYQNPsnDtTdbCCN9DC0DDXXPZgopGOIaP4NyNR0ttaNfNUQrrO18rtODtB4NExDQx4WDgwJLCw8VfAGMACgfV9v2MKzMiCCIYjsFIAvyEoajg7oXD6IYxjWNY%2Bg09tkCLKgiSdITW3tPS6QuAw7ieC0eghNMJRlHoeRpAIYx%2BDLVudH05tzLYOs1AI3SjEb2Qy1U7tdJMTsDOUww9Lbej4j0QezOUiwKM1IslWVq1oxt2yqMN05SuN2zAMW2yPX83oQLghCPGCXDzLwnVaPMvUgJIZh/OOj2guOZj%2BBU44AJz%2BKNS0cCtpAs%2BOu2VdVG1bTte1dQdx0QEgxMlgyV0LsQnPKIY7RCAgqDOv9vA3ed5bpJv4S0Dve/j4fqC3fQ92PY6R93d9QqX/v4NL0C69a0EqjVHUfAlVeD8EFqIcQotQHixUOoNG0tZZGBQArSwStogqx7OrTWnAtoJwjpeMIZ9t67w/twau3Iya8GdHTRIrMk4cHKqQa%2BNVODYH/svR4Gcs45zzsgAuTdi71SnDYbYpcGStUrtXfaiwbivEGD2Uqg9eAjzHuDSethp4126qQPqjdm6t3bp3fwPc%2B6BAUaCFOE9sEz1rnQswFj1pWM0XXUgdJEJ60kEAA)

klenze: You are correct. 

The [standard implementation](https://en.cppreference.com/w/cpp/algorithm/accumulate) apparently looks like this:
```
template<class InputIt, class T, class BinaryOperation> 
constexpr // since C++20 
T accumulate(InputIt first, InputIt last, T init, BinaryOperation op) 
{     
       for (; first != last; ++first)
           init = op(std::move(init), *first); // std::move since C++20
       return init; 
}
```

So the first argument to op is something which gets recycled. If we return the first argument by rvalue reference, the assignment becomes effectively:
```
      init=std::move(init);
```
which will in fact clear init. 

I think readability (as well as number of deep/shallow copies) could probably be improved by just using a range loop:
```
TString reaction;
for (auto& t: tracks)
{
    ...
    reaction += "_" + TString::Itoa(t->GetPdgCode(), 10);
}
```
Instead of the verbose find call, one might perhaps use set::count?

YanzhaoW: I actually haven't yet checked this code. But I don't see anything wrong in your suggestion.

Maybe also leave it as close to its original version as possible? In the recent future I will probably do some refactoring for the neuland code. Then I will change it to a better form given that we still need this source file.

klenze: ok. fwiw, here was the version I would have used:
```
                const std::set<int> filteredPdgCodes = { 22, 111 };
                TString reaction;
                for (auto& t : tracks)
                    if (!filteredPdgCodes.count(t->GetPdgCode()))
                        reaction += "_" + TString::Itoa(t->GetPdgCode(), 10);
```

klenze: Looks like the lambda with the first argument being an rvalue reference does not compile on C++17. 

The signature is now ``[&](TString s, const R3BMCTrack* b) -> TString`` which should compile everywhere.

jose-luis-rs:
Please, solve the conflicts for merging.

jose-luis-rs:
> @kresan @jose-luis-rs: Is there some mechanism to put larger files for retrieval on the web? I previously tried pastebin sites, but they refuse to work with a megabyte of compile logs, and clearly my filtering script is not working here.

I don't know, sorry!

jose-luis-rs:
Thanks

PR #746, opened 2023-01-09
Update year for all license headers

jose-luis-rs:


klenze:
While I find the copyright banner somewhat distasteful in the first place (even when used for the purpose of copyleft), allow me to nitpick. I think simply changing the year is not correct.

From my understanding, the copyright year should either be:
* The year of first publication
* A comma separated list of years in which revisions were made
* A range covering the time span in which revisions were made 

Generally speaking, the copyright year is given so that the public can figure out if a work is in the public domain or not. 

The relevant authority on copyright law, GNU, [states](https://www.gnu.org/licenses/gpl-howto.en.html) the following:
> The copyright notice should include the year in which you finished preparing the release (so if you finished it in 1998 but didn't post it until 1999, use 1998).  You should add the proper year for each past release; for example, “Copyright 1998, 1999 Terry Jones” if some releases were finished in 1998 and some were finished in 1999. [...]
> For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”) if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage.

PR #748, opened 2023-01-10
Renamed CalifaHit to CalifaCluster

jose-luis-rs:
Following #720

### DIFF
--- a/analysis/offline/R3BTrackS454.cxx
+++ b/analysis/offline/R3BTrackS454.cxx
@@ -853,7 +853,7 @@ void R3BTrackS454::Exec(Option_t* option)
         return;
     }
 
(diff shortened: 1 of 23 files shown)


ryotani:
What's the difference from #720? Is this basically a squash merge + some additional changes?

jose-luis-rs:
The main difference is the use of Cluster instead of Hit and also R3BCalifaHitPar was renamed to R3BCalifaClusterPar.

jose-luis-rs:
At the moment R3BCalifaHitPar (R3BCalifaClusterPar) is not used for analysis, so you don't need to change the names of your parameters. Really, it is not loaded by the container factory R3BCalifaContFact ...

ryotani:
Ah, right. Thank you!

PR #749, opened 2023-01-10
Update R3BFiberMAPMTCal2Hit.cxx

AndreaLagni:
removed line 522, tof condition not working

### DIFF
--- a/fiber/mapmt/R3BFiberMAPMTCal2Hit.cxx
+++ b/fiber/mapmt/R3BFiberMAPMTCal2Hit.cxx
@@ -519,11 +519,8 @@ void R3BFiberMAPMTCal2Hit::Exec(Option_t* option)
 
                     if (!fIsCalibrator)
                     {
-                        if (tof >= ftofmin && tof <= ftofmax)
-                        {
                             new ((*fHitItems)[fHitItems->GetEntriesFast()]) R3BFiberMAPMTHitData(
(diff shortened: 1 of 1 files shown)


ajedele:
I would prefer if this pull request was paused so that we can get feedback from the calibration procedure developers on why that step was implemented. In the meantime, that line can be commented out of rewritten.

jose-luis-rs:
You could open the window, now these parameters are set to:
    , ftofmin(-1000)
    , ftofmax(1000)
you could change these values from your analysis macro using the function:
SetTimeWindow(Double_t tmin, Double_t tmax)

I guess the fibers are not calibrated and for this reason it fails, this window should be used after calibration to select good hits and remove the background.

ajedele:
The fibers are calibrated and the hit-level data should be available. I'm not sure why this cannot wait until Enis returns from vacation next Monday. I'm also not convinced this is the newest code. Enis has made several modifications that may not have been pulled yet.

jose-luis-rs:
Yes, I can confirm that there are no PRs related to the fibers.

AndreaLagni:
We are working on MDF and we saw that the actual version of the class is originating a lot of multiplicity 0 events, due to the tof condition in line 522. So, that's why we modified the class. Then,  I thought that this was an updated class and not old one. I will discuss with Enis once he is back.

PR #751, opened 2023-01-10
Use of R3BStack instead of FairStack in gconfig/g4Config.C

jose-luis-rs:
Note that some tests failed during the last PRs and we were not alerted by github!!!


jose-luis-rs:
Ready!

PR #752, opened 2023-01-11
CALIFA: deleted old root geometries

jose-luis-rs:
testCalifaSimulation.C: Update name for the Califa root geometry


csuerder:
For s467 we are using the 2020 geometry and the files califa_2020.geo.root and califa_2020_s444.geo.root. They are deleted with others and only two geometry versions are left for CALIFA. Will there be a substitute or how can we use our geometry? Also if I understand this correctly the R3BCalifaGeometry::Init(Int_t version) will now have issues. Can you comment on that?

jose-luis-rs:
Yes, the new versions are:
- califa_2020_s444.geo.root ->  califa_v2019.11.geo.root
- califa_2020.geo.root -> califa_v2021.3.geo.root

Then the full califa geometry corresponds to califa_full.geo.root (full barrel + iphos)

csuerder:
Thanks for the answer. We will wait for the outcome of the discussion. And for now I can not see the new geometry versions in dev, I assume they will be uploaded, soon?

jose-luis-rs:
Now the root geometries are generated automatically during the R3BRoot compilation.

PR #754, opened 2023-01-16
Incoming detector updates

nmozumdar:


nmozumdar:
First Commit for Incoming detectors with new reconstruction  method for finding hit in Los and introduced multi-hit capability of Musli. 
The parameter finders are currently being worked on.

jose-luis-rs:
I think that this PR needs the acceptance of #753

jose-luis-rs:
Please, try to rebase r3b/dev to see what happens.

nmozumdar:
Okay thank you very much, it seems to be working now. Please review the code and if it seems ok feel free to merge it. I'll make a separate PR for the FRSdata tasks.

los/calib/R3BLosTCal2Hit.cxx
```diff
@@ -0,0 +1,496 @@
...
+ *                    copied verbatim in the file "LICENSE".                  *
+ *                                                                            *
+ * In applying this license GSI does not waive the privileges and immunities  *
+ * granted to it by virtue of its status as an Intergovernmental Organization *
+ * or submit itself to any jurisdiction.                                      *
+ ******************************************************************************/
+
+// -------------------------------------------------------------
+// -----                  R3BLosTCal2Hit                   -----
+// -----          Created Mar 10th 2016 by R.Plag          -----
```
jose-luis-rs, line 16: Update author and dates

ajedele:
I have reviewed the functionality of the LOS code. The functionalities are consistent with the previous ones.

For the R3BLosProvideTStart code, the main difference is adding the TriggerData, which is consistent with any detector now using the T3 trigger. This is the 1st experiment where T3 was added for LOS.

For the R3BLosHitPar code, the main difference is adding the parameter fNumParamsTamexLE and fNumParamsTamexTE, and placing the params into a TArrayF. There are added LOG(INFO) statements, which is always good.

For the R3BLosTCal2Hit calibration, there are a lot of changes. 
The changes in the init seems to be mostly to make the code compatible with the newest version of r3broot and with the new params structure. 
The calculation of the fine time correction now includes the new TimeStitcher function, to make the calculation consistent across all detectors.
The TriggerItems is also added to separate the triggers from the raw data. This is consistent with the structure of the ToFD, which uses the same electronics.
The biggest change is adding the loop over the number of hits. The filtering step to verify if an event is a good event and is in the window of interest is still consistent with the previous method. The position and average time calibration is exactly the same procedure as the previous code - only modified for multi-hits. The same holds for the saturation correction method and the walk correction.

From the perspective of the functionality of the detector, I agree with the changes implemented and would recommend accepting the LOS contribution of the pull request.

For the MUSLI changes, I would recommend a review from @audreychatillon and the SOFIA group.

musli/pars/R3BMusliCalPar.h
```diff
@@ -76,15 +79,23 @@ class R3BMusliCalPar : public FairParGenericSet
         int index = (signal - 1) * fNumParamsPosFit + indexpar - 1;
         fPosCalParams->AddAt(val, index);
     }
+    void SetMultHitParams(Float_t val, Int_t signal, Int_t indexpar)
+    {
+        // 1-base
```
jose-luis-rs, line 84: 0-base ??

jose-luis-rs: In the file R3BMusliHitPar.h, type and deg should be 1-base and the index 0-base because it is used by a TArrayF.
0-base refers to the index variable like you did for the class R3BLosHitPar.h

    void SetEcorrBetaHitParams(Double_t val, Int_t type, Int_t deg)
    {
        // 0 - based
        int index = (type - 1) * fNumParamsEcorrBetaFit + deg - 1;
        fEcorrBetaHitParams->AddAt(val, index);
    }
    void SetZHitParams(Double_t val, Int_t type, Int_t deg)
    {
        // 0 - based
        int index = (type - 1) * fNumParamsZFit + deg - 1;
        fZHitParams->AddAt(val, type - 1);
    }
Maybe there is a mistake in "fZHitParams->AddAt(val, type - 1);" =>    type-1 --> index ??
Please, could you check if there are mistakes in the classes calling these parameter containers?

In R3BMusliCalPar.h signal and indexpar -> 1-base and index --> 0-base

    void SetEneParams(Float_t val, Int_t signal, Int_t indexpar)
    {
        // 1-base  ---> 0-base
        int index = (signal - 1) * fNumParamsEneFit + indexpar - 1;
        fEneCalParams->AddAt(val, index);
    }
    void SetPosParams(Float_t val, Int_t signal, Int_t indexpar)
    {
        // 1-base  ---> 0-base
        int index = (signal - 1) * fNumParamsPosFit + indexpar - 1;
        fPosCalParams->AddAt(val, index);
    }
Please, check that everything works as expected because these classes were developed during the beam time and the probability of mistakes is high.

In the case of R3BLosHitPar.h, pmt and indexpar should be 1-base but seems like indexpar is 1-base and pmt 0-base (see below)

    void SetLEMatchParams(Float_t val, Int_t pmt, Int_t indexpar)
    {
        // 0-base
        int index = pmt * fNumParamsTamexLE + indexpar - 1;
        fLEMatchParams->AddAt(val, index);
    }
    void SetTEMatchParams(Float_t val, Int_t pmt, Int_t indexpar)
    {
        // 0-base
        int index = pmt * fNumParamsTamexTE + indexpar - 1;
        fTEMatchParams->AddAt(val, index);
    }
If you have time, please, check that this part also works correctly.

You can do it when you wish, but if other collaborators need this code to continue with the analysis I think that you could do these corrections over the current PR.

nmozumdar: Done, made things consistent with respective parameter finders and data structures. Note that most of these param setters are not called by anything within R3Broot because the parameter finders do not exist, for eg. there is no MusliCal2HitPar, and I just find the parameters by myself.

Please do let me know if there are any more corrections

jose-luis-rs (approved):

klenze:
NB: this commit is credited to "A land user who did not set git config --local <land@gsi.de>" in git log.

nmozumdar:
Sorry I fixed my local config now and my latest PR's commit should have an author. I feel hesitant to set these things especially in the Gsi PCs since I am not extremely familiar with Git.

PR #768, opened 2023-01-27
Update FrsData tasks for s509

nmozumdar:
New tasks for Incoming ID analysis to correct for loss of statistics problem in S2  for s509 experiment due to both PMTs not being present in ~20-25% events. 
The R3BIncomingBeta509 does most of the things as the older existing task to keep it backwards compatible. In addition to the fixing of the problem mentioned above it is also multi-hit capable when using Los Z (check comments in the task).

In addition made R3BAnalysisIncomingID509 consistent with the above problem and improvements when selecting Los charge. It is also mostly backwards compatible except when using Los charge expectation is that LosProvideTstart is use for beta calculation and one has to at least run new Task R3BLosMapped2TCal even if they wish to use the older Los tasks. If other detectors are used for Z measurement then its fully backwards compatible.

Of course one has the choice to ignore these new tasks and use the older tasks since I didn't touch them. These new tasks are however mandatory for s509 and heavily recommended for s522.

### DIFF
--- a/analysis/offline/R3BAnalysisIncomingID.cxx
+++ b/analysis/offline/R3BAnalysisIncomingID.cxx
@@ -12,7 +12,7 @@
  ******************************************************************************/
 
 // ----------------------------------------------------------------
-// -----       R3BAnalysisIncomingID source file              -----
+// -----             R3BAnalysisIncomingID                    -----
 // -----     Created 01/11/21 by M. Feijoo Fontan             -----
 // ----------------------------------------------------------------
 
@@ -27,6 +27,7 @@
 #include "FairRuntimeDb.h"
 
 #include "R3BAnalysisIncomingID.h"
+#include "R3BCoarseTimeStitch.h"
 #include "R3BEventHeader.h"
 #include "R3BFrsData.h"
 #include "R3BIncomingIDPar.h"
@@ -34,6 +35,7 @@
 #include "R3BLosCalData.h"
 #include "R3BLosHitData.h"
 #include "R3BLosMappedData.h"
+#include "R3BLosTCalData.h"
 #include "R3BMusicHitData.h"
 #include "R3BMusicHitPar.h"
 #include "R3BMusliHitData.h"
@@ -49,10 +51,10 @@ R3BAnalysisIncomingID::R3BAnalysisIncomingID()
 
 R3BAnalysisIncomingID::R3BAnalysisIncomingID(const char* name, Int_t iVerbose)
     : FairTask(name, iVerbose)
-    , fHeader(NULL)
     , fHitItemsMus(NULL)
     , fHitItemsMusli(NULL)
     , fHitLos(NULL)
+    , fTriggerLos(NULL)
     , fHitPspx1_x(NULL)
     , fHitPspx1_y(NULL)
     , fFrsDataCA(NULL)
@@ -68,6 +70,8 @@ R3BAnalysisIncomingID::R3BAnalysisIncomingID(const char* name, Int_t iVerbose)
     , fNumDet(1)
     , fUseLOS(kFALSE)
     , fUsePspx1(kTRUE)
+    , fUseTref(kFALSE)
+    , fTimeStitch(nullptr)
     , fCutS2(NULL)
     , fCutCave(NULL)
 {
@@ -150,6 +154,13 @@ InitStatus R3BAnalysisIncomingID::Init()
     fHitLos = (TClonesArray*)mgr->GetObject("LosHit");
     R3BLOG_IF(warn, !fHitLos, "LosHit not found");
 
+    if (fUseTref)
+    {
+        // Get access to trigger data of the LOS
+        fTriggerLos = (TClonesArray*)mgr->GetObject("LosTriggerTCal");
+        R3BLOG_IF(warn, !fTriggerLos, "LosTriggerTCal not found");
+    }
+
     // Get access to hit data of PSPX1
     fHitPspx1_x = (TClonesArray*)mgr->GetObject("Pspx1_xHit");
     R3BLOG_IF(warn, !fHitPspx1_x, "Pspx1_xHit not found");
@@ -163,6 +174,9 @@ InitStatus R3BAnalysisIncomingID::Init()
         fFrsDataCA = new TClonesArray("R3BFrsData");
         mgr->Register("FrsData", "Analysis FRS", fFrsDataCA, !fOnline);
     }
+
+    // Definition of a time stich object to correlate times coming from different systems
+    fTimeStitch = new R3BCoarseTimeStitch();
 
     SetParameter();
     return kSUCCESS;
@@ -210,6 +224,7 @@ void R3BAnalysisIncomingID::Exec(Option_t* option)
     Double_t Zlos[fNumDet];
     UInt_t nHits = 0;
     Double_t posLosX_cm[fNumDet];
+    Double_t trigTimeV[fNumDet];
     Double_t Gamma_m1 = 0., Brho_m1 = 0., AoQ_m1 = 0.;
     Double_t AoQ_m1_corr = 0.;
     Int_t multLos[fNumDet];
@@ -220,6 +235,20 @@ void R3BAnalysisIncomingID::Exec(Option_t* option)
         multLos[i] = 0;
         Zlos[i] = 0.;
         posLosX_cm[i] = 0.;
+        trigTimeV[i] = 0.;
+    }
+    // --- read Trigger data from LOS --- //
+    if (fTriggerLos && fTriggerLos->GetEntriesFast() > 0)
+    {
+        Int_t numDet = 1;
+        Int_t tHits = fTriggerLos->GetEntriesFast();
+        for (Int_t ihit = 0; ihit < tHits; ihit++)
+        {
+            R3BLosTCalData* hittcal = (R3BLosTCalData*)fTriggerLos->At(ihit);
+            numDet = hittcal->GetDetector();
+            if (hittcal->GetType() == 0)
+                trigTimeV[numDet - 1] = hittcal->GetRawTimeNs();
+        } // --- end of loop over hit data --- //
     }
 
     // --- read hit from LOS data --- //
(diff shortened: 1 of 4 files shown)


jose-luis-rs:
Hello

What is the difference with respect to R3BAnalysisIncomingID and  R3BIncomingBeta?

If there are small differences one can use the header->GetExpId() getter to get the expID and use it to include specific analysis conditions, see for instance R3BTwimCal2Hit.cxx 
https://github.com/R3BRootGroup/R3BRoot/blob/578e697bf2dec7cf32e5f549eaa8c4d39625384c/twim/calibration/R3BTwimCal2Hit.cxx#L327

nmozumdar:
Hi for R3BIncomingbeta509 there are already experiment specific conditions where I read the experiment ID from the event header so other experiments should in principle get the exact same results as before if they do nothing i.e. it is fully backward compatible.
 
For R3BAnalysisIncomingID509, only when one uses los Z  for charge there are some requirements but I can also make that experiment specific or set the fUseTref bool condition as in R3BIncomingBeta. and then it should also be fully compatible with previous version.

If you agree I can make the changes and push them as the original versions, I just wanted to be safe so no mistakes creep up on other experiments. (but maybe this can be checked when reviewing the PR and feedback from the other experiments using this task)

Also I don't want to use the R3BTwimCal2Hit approach of multiple member functions since it will result in a lot of repeated code.

jose-luis-rs:
If this is for test, we can accept it.

jose-luis-rs:
@klenze do you agree?

klenze:
> Also I don't want to use the R3BTwimCal2Hit approach of multiple member functions since it will result in a lot of repeated code.

Agreed. Preferable, experiment specific stuff should be explicitly set using parameters or the macro itself. 

> If you agree I can make the changes and push them as the original versions, I just wanted to be safe so no mistakes creep up on other experiments. 

In a perfect world, the other experiments would have CI tests in place to detect any effects on their analysis. In our fallen world, the reason most experiments will not be affected is simply because they long ago forked and are not following the dev branch any more. 

I think that creating an experiment-specific class should be limited to the most extreme cases. For [R3BS515PspxReader](https://github.com/R3BRootGroup/R3BRoot/blob/dev/r3bsource/pspx/R3BS515PspxReader.cxx), Ivana had to insert code to buffer and delay PSP events due to readout troubles. This was the only way to make sense of the data in that experiment, but nothing we will ever need again (due to the DAQ group subsequently adding safeguards against such things). From what I gather, your case is nothing like this. 

Make a good effort to not break the behavior of the code for previous experiments (unless it is buggy in the first place) without consulting them, and add your code to the existing class. 

Creating a new file also makes it hard to review your changes: I would have to checkout your PR and manually do a diff between the two files. 

Without doing that, some general remarks (which may be stuff written by you or stuff you copied over):
```
    , fHeader(NULL)
```
Please initialize members to fixed values in the header instead:
```
protected:
    MyHeaderType* fHeader{};
```

```
    , fP2(-2.87635e7)
```
Please try to avoid to putting specific values which will differ among experiments into the code. If you don't want to deal with our parameter format, have people set them using a setter. If fP0, fP1 and fP2 were to form a spatial vector, consider using TVector3 or something. 

```
    fToFoffset = new TArrayF(fNumDet);
```

use std::vector.

fNumDet seems more like fMaxMultiplicity to me?

By my count, Exec is 228 lines long. That is roughly 180 lines too long. 

One way to shorten the code would be the following:
* Several detectors (e.g. Los, Pspx, Music) implement charge determination. 
* An abstract base class R3BIChargeMeasurement with a method ``virtual double GetZ()=0`` is created
* All of the Hit level data of the detectors in question implements this interface (e.g., is derived from R3BIChargeMeasurement). 
* Instead of specifying which detectors to use, the user provides a ``std::vector<std::string>`` of TClonesArray names of such detectors. 
* In Init, these get converted into a ``std::vector<TClonesArray*>``.
* In Exec, Z is determined by some defined algorithm (e.g. "take the Z from the first detector in the vector offering a valid measurement").

Unfortunately, this is the real world, not my to example, so the job will not be as easy as just determining Z. Still, I can't help but notice that the handling of Music and Musli is very similar and wonder if we might not be better of if they used the same hit level data structure. 

```
        ZPsp = en_pspx;
```
naively, I would have assumed that (en_pspx+en_pspy)/2 might offer a better resolution.

```
Double_t time = fTimeStitch->GetTime(hittcal->GetTime() - trigTimeV[numDet - 1], "vftx", "vftx");
if (time == fHeader->GetTStart())
```
Please avoid comparing (what appears to be) arbitrary floating point values using equality. 

```
            if (num_tof_candidates == 0 && fHeader->GetExpId() == 509)
```
Please use a flag instead: `` if (num_tof_candidates == 0 && fSomeFlagDescribingTheNewBehavior)``

```
                            Beta_m1 = Velo_m1 / 0.299792458;
```
I would prefer to refer to fundamental constants through some const symbol -- what if the speed of light were to change during the operational phase of FAIR? :-)

Using [CLHEP](https://gitlab.cern.ch/CLHEP/CLHEP/-/tree/develop/Units/Units), `` c_light / (m/ns)`` should do the trick. I am sure ROOT has a way to express the same. 

```
void R3BIncomingBeta509::FinishEvent()
{
    if (fHitSci2)
    {
        fHitSci2->Clear();
    }
    if (fHitLos)
    {
        fHitLos->Clear();
    }
}
```
What business do you have clearing TCAs written by other tasks? You task owns fFrsDataCA, let's worry about that only. 

Even more severe in the other Exec method. (Why are there two classes in the first place?)

Probably, most of the stuff is I mentioned is stuff you did not write but only copied from existing code. Feel free to ignore that.

nmozumdar:
Thanks a lot for looking into the code and for the suggestions. I have now pushed it as an update to the existing task.

> Without doing that, some general remarks (which may be stuff written by you or stuff you copied over):
> 
> ```
>     , fHeader(NULL)
> ```
> 
> Please initialize members to fixed values in the header instead:
> 
> ```
> protected:
>     MyHeaderType* fHeader{};
> ```

Done (hopefully correctly)

> ```
>     , fP2(-2.87635e7)
> ```
> 
> Please try to avoid to putting specific values which will differ among experiments into the code. If you don't want to deal with our parameter format, have people set them using a setter. If fP0, fP1 and fP2 were to form a spatial vector, consider using TVector3 or something.

Those values are not initialized by me. I use my own parameter file. But I leave it as it is since some other experiment might not have a parameter file but rather depend on these default parameters. (I don't know honestly if this is the case.)
fP0, fP1 and fP2 are charge calibration parameters, quite redundant tbh since this calibration is done at the hit level of the individual detectors. (In fact they are not even used properly in the task)

> ```
>     fToFoffset = new TArrayF(fNumDet);
> ```
> 
> use std::vector.

If I change this here I will have to change the types in the parameter read/writers to keep things consistent.

> fNumDet seems more like fMaxMultiplicity to me?

No, different experiments have different number of LOS and FRS start detectors, so one can take tof between different combination of detectors which is why one has the choice to store the detector IDs. It does not affect my experiment since I have only 1 of each. Also looping over the same number of detectors is not correct as (I think) there are experiments with 2 LOS and 1 FRS detectors etc. 

> By my count, Exec is 228 lines long. That is roughly 180 lines too long.
> 
> One way to shorten the code would be the following:
> 
> * Several detectors (e.g. Los, Pspx, Music) implement charge determination.
> * An abstract base class R3BIChargeMeasurement with a method `virtual double GetZ()=0` is created
> * All of the Hit level data of the detectors in question implements this interface (e.g., is derived from R3BIChargeMeasurement).
> * Instead of specifying which detectors to use, the user provides a `std::vector<std::string>` of TClonesArray names of such detectors.
> * In Init, these get converted into a `std::vector<TClonesArray*>`.
> * In Exec, Z is determined by some defined algorithm (e.g. "take the Z from the first detector in the vector offering a valid measurement").

This seems like an excellent idea and I would love to implement it but it would take some time since it requires the understanding of the working of all the charge detectors using the task,
   
> Unfortunately, this is the real world, not my to example, so the job will not be as easy as just determining Z. Still, I can't help but notice that the handling of Music and Musli is very similar and wonder if we might not be better of if they used the same hit level data structure.

This is something I agree with, Music and Musli can indeed have a common data structure even if there are some fundamental differences between the two detection systems. However at least from the Musli side I can say (without going into details) that this needs a consensus on what type of data is finally stored at hit level.   

> ```
>         ZPsp = en_pspx;
> ```
> 
> naively, I would have assumed that (en_pspx+en_pspy)/2 might offer a better resolution.

Makes sense but I don't know enough about these so I leave it as it is.
 
> ```
> Double_t time = fTimeStitch->GetTime(hittcal->GetTime() - trigTimeV[numDet - 1], "vftx", "vftx");
> if (time == fHeader->GetTStart())
> ```
> 
> Please avoid comparing (what appears to be) arbitrary floating point values using equality.

They are not arbitrary, that equation is the exact way the TStart() is calculated. I do this because in the R3BIncomingBeta when using trigger corrected times, the tof is measured using the TStart() time, therefore when saving the Z in this separate task we must select the Z from the hit corresponding to the TStart hit and this equality is the only way to do that. Also if using LOS Z the better option is to only use the R3BIncomingBeta task only and saving the data the (my comments in the task) 

> ```
>             if (num_tof_candidates == 0 && fHeader->GetExpId() == 509)
> ```
> 
> Please use a flag instead: ` if (num_tof_candidates == 0 && fSomeFlagDescribingTheNewBehavior)`

` num_tof_candidates == 0` IS the flag for the problem. For a LOS hit looping over all hits of S2, if no S2 hit satisfying the beta condition is found then this is triggered. the ExpID is just to ensure that this is done only for s509.

> ```
>                             Beta_m1 = Velo_m1 / 0.299792458;
> ```
> 
> I would prefer to refer to fundamental constants through some const symbol -- what if the speed of light were to change during the operational phase of FAIR? :-)
> 
> Using [CLHEP](https://gitlab.cern.ch/CLHEP/CLHEP/-/tree/develop/Units/Units), ` c_light / (m/ns)` should do the trick. I am sure ROOT has a way to express the same.

The best I found is the `TMath::C()` which gives the values in m/s and I just divide by 10^9 explicitly.

> ```
> void R3BIncomingBeta509::FinishEvent()
> {
>     if (fHitSci2)
>     {
>         fHitSci2->Clear();
>     }
>     if (fHitLos)
>     {
>         fHitLos->Clear();
>     }
> }
> ```
> 
> What business do you have clearing TCAs written by other tasks? You task owns fFrsDataCA, let's worry about that only.
> 
> Even more severe in the other Exec method. (Why are there two classes in the first place?)

I can remove this but from my naïve viewpoint doesn't the Clear() method just deallocate the memory for the TCA pointers fHitSci2, fHitLos ... which was initialized in the task. Is this bad?

There are two classes because Musli (and Music also i think) need beta correction in their charge values to write to the hit level. So the beta is first calculated, applied to charge calibration in Musli (Music) and then that charge is written in FRSdata with the second task. The alternative I would think is to make one FRS task, read Cal data of Musli (Music) and do the hit level calibration + Frs data calculation in there which is obviously not ideal since its better to have a separate calibration task for individual detectors.

jose-luis-rs (approved):

PR #773, opened 2023-02-01
Added offset of 2cm on the Y axis of GLAD

jose-luis-rs:
Clang-format as well

### DIFF
--- a/passive/R3BGladMagnet.cxx
+++ b/passive/R3BGladMagnet.cxx
@@ -15,10 +15,10 @@
 #include "FairLogger.h"
 
 // NOTE: as for now, these values are the same used
-//       for the geometry creation (v17).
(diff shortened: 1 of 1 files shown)


klenze:
Could we please clang-format all files and then enforce clang-format for all files? This "if you change it you are responsible for formatting all of the file" is silly. See also #736.

jose-luis-rs:
Hello

I just ran the current clang-format file:
. util/clang-format-all.sh

I know that you have an update for the file .clang-format, but did you check it on different systems (Linux, debian, mac, centos,...) or only on GSI machines?
I would like to be sure that all our collaborators can use this script before updating it.

As you can see in other emails, the last modifications are causing/adding analysis stoppers to other collaborators.

jose-luis-rs:
Ready!

PR #776, opened 2023-02-02
Added automatic root geometry generation of ALPIDE v23

jose-luis-rs:


YanzhaoW:
The generation of geometry files during cmake configuration should be disabled by default, in my opinion. For most of R3BRoot users, I would guess, they just need one or few. (And my output of cmake configuration is littered with geometry generation failures in my machine.) I would suggest to add some cache variables that can be set to initiate the geometry generation.

We could discuss about this in the analysis meeting.

jose-luis-rs:
I agree, this is a basic change. But meanwhile I would like to know about the problems. 
Could you upload the log file? or copy&paste?
Thanks in advance

jose-luis-rs:
At the GSI machines one needs:

export ROOTSYS=$SIMPATH
export PATH=$ROOTSYS/bin':'$PATH
export LD_LIBRARY_PATH=$ROOTSYS/lib/root':'$LD_LIBRARY_PATH

to avoid the problems with the generation of root geometries.

YanzhaoW:
> 

Do you mean the output from cmake? If so, please check the file below.
[log.txt.zip](https://github.com/R3BRootGroup/R3BRoot/files/10572258/log.txt.zip)

jose-luis-rs:
Yes, but in the attached file everything is OK. Please, delete first the root geometries in the folder geometry and then
- cd build
- rm CMakeCache.txt
- rerun the cmake
then we can see the problems.

But in the GSI computers, if you add these lines to your environment:

export ROOTSYS=$SIMPATH
export PATH=$ROOTSYS/bin':'$PATH
export LD_LIBRARY_PATH=$ROOTSYS/lib/root':'$LD_LIBRARY_PATH

the automatic generation should works properly. For a strange reason it does not find some root tools...

klenze:
Some more nitpicks about autogenerating geometry files:

* The input files should not be in macros. I support the decision to move tests from the macros directory, but this is much more helpful if we manage to avoid new dependencies to subrepositories. Or course, if the contents of macros were to be merged into the R3BRoot repository, this issue would be moot. 
* cmake should do the bare minimum to figure out what parameters need to be set for the build process. Running root to create a geometry file from a macro should not part of that process any more than running g++ to compile sources (apart from tests if the compiler works). The best way to implement this would be to use [add_custom_command](https://cmake.org/cmake/help/latest/command/add_custom_command.html#command:add_custom_command). 
* In the geometries are autogenerated, they should not be placed in the source directory, but in the build directory, and should be loaded from selfsame. If the geo file is not autogenerated yet, it should instead be copied from the source directory. 

Also, for CALIFA specifically, I maintain that the following may be undesirable:
```
(for experimental setup)
    f = new TFile(geoPath, "READ");
    TGeoVolume* v = dynamic_cast<TGeoVolume*>(f->Get("TOP"));
    gGeoManager->SetTopVolume(v);
```

From the starting letter of gGeoManager, I assume it is effectively a global variable. Now I don't claim to understand what [SetTopVolume](https://root.cern.ch/doc/master/TGeoManager_8cxx_source.html#l03409) does exactly, but from my point of view it looks like it is mostly replacing whatever was the top volume before (rather then merging more volumes onto the top volume by path). This would imply that the R3BCalifaGeometry approach works only if califa is the only detector doing it.
A better approach would be to store with every geometry file an independently readable file of califa positions. (preferably human-readable text).

jose-luis-rs:
> Some more nitpicks about autogenerating geometry files:
> 
>     * The input files should not be in macros. I support the decision to move tests from the macros directory, but this is much more helpful if we manage to avoid new dependencies to subrepositories. Or course, if the contents of macros were to be merged into the R3BRoot repository, this issue would be moot.

I think that at the end we will merge everything after cleaning the repo macros. 

> 
>     * cmake should do the bare minimum to figure out what parameters need to be set for the build process. Running root to create a geometry file from a macro should not part of that process any more than running g++ to compile sources (apart from tests if the compiler works). The best way to implement this would be to use [add_custom_command](https://cmake.org/cmake/help/latest/command/add_custom_command.html#command:add_custom_command).

I am not sure how to use the command "add_custom_command" to run a macro, but we also need to redo all the cmake files following the new implementation of FairRoot. I guess that we also have to change all the configuration files used for the cmakes.

> 
>     * In the geometries are autogenerated, they should not be placed in the source directory, but in the build directory, and should be loaded from selfsame. If the geo file is not autogenerated yet, it should instead be copied from the source directory.

This is possible, but then the users must define the GEOMPATH properly in their macros for simulation, I am pretty sure that this will cause problems handling the code.


> 
> 
> Also, for CALIFA specifically, I maintain that the following may be undesirable:
> 
> ```
> (for experimental setup)
>     f = new TFile(geoPath, "READ");
>     TGeoVolume* v = dynamic_cast<TGeoVolume*>(f->Get("TOP"));
>     gGeoManager->SetTopVolume(v);
> ```
> 
> From the starting letter of gGeoManager, I assume it is effectively a global variable. Now I don't claim to understand what [SetTopVolume](https://root.cern.ch/doc/master/TGeoManager_8cxx_source.html#l03409) does exactly, but from my point of view it looks like it is mostly replacing whatever was the top volume before (rather then merging more volumes onto the top volume by path). This would imply that the R3BCalifaGeometry approach works only if califa is the only detector doing it. A better approach would be to store with every geometry file an independently readable file of califa positions. (preferably human-readable text).

At the moment this was only tested with FairSoft-jun19p2 and FairRoot-v18.2.1, using CALIFA+ALPIDE and everything was OK.
With the new versions I don't know yet, really we are doing many changes in the code that also depend on FairSoft/FairRoot, so we have to see from now how their modifications affect to our simulations/analysis. It seems that the autogeneration of root geometries works fine on Ubuntu, but fails in Debian and Centos...

> 
> Edit: also, it might be a good idea to store checksums of geometry files and compare them to previous values. Of course, floating point numbers compressed to binary make this so much more interesting. The best way to do it would perhaps be a volume compare function which loads two files worth of geometries, recursively traverses all the pathes and test if the points defining each object are numerically close to each other.

You mean the comparison of two root geometry files? Initially your system will produce one root file by default, then you could compare other files modified by you with the default one. But I don't see how this could solve the current problems! 
The idea of having the root files under git was that the root geometries are produced and tested by the detector experts, and then submitted to the r3broot repository for their use. Now we are generating most of them in automatic mode and this is causing alterations in the position of some components. More investigations are needed ...........................

jose-luis-rs:
With this PR all checkings are in green:
Screenshot from 2023-02-10 13-59-06

PR #778, opened 2023-02-03
Added BUILD_GEOMETRY to the CMake to disable the generation of root g…

jose-luis-rs:
…eometries


jose-luis-rs:
By default -DBUILD_GEOMETRY should be ON
if someone does not want to generate the root geometries, execute the cmake as:
cmake ../R3BRoot -DBUILD_GEOMETRY=OFF
As requested by #776, but considering that the most general case is the analysis of a full experiment.

PR #788, opened 2023-02-06
Fix bugs with out-of-boundary acess of std::vector in cosmic tracker

YanzhaoW:
Fixes two bugs discovered by Philip (@klenze ) and Igor(@igasparic ).

This PR is just a quick fix. As as been shown many times, using raw index of a container is very likely to cause issues. I will probably rewrite the whole NeulandCosmicTracker later when I have time.

Hi, @vadimr3b, is there any way I can contact you these days?

### DIFF
--- a/neuland/calibration/R3BNeulandCosmicTracker.cxx
+++ b/neuland/calibration/R3BNeulandCosmicTracker.cxx
@@ -67,12 +67,13 @@ namespace Neuland
         }
 
         CosmicTracker::CosmicTracker()
-            : fFit("CosmicTracker:fFit", "pol1")
+            : fDistances{ 0. }
+            , fFit("CosmicTracker:fFit", "pol1")
         {
             fDistances.reserve(MaxNumberOfPlanes);
-            for (auto p = 0; p < MaxNumberOfPlanes; ++p)
+            while (fDistances.size() < MaxNumberOfPlanes)
             {
-                fDistances[p] = p * BarSize_Z;
+                fDistances.push_back(fDistances.back() + BarSize_Z);
             }
             fTrack.Interactions.reserve(256);
             fBarIDs.reserve(256);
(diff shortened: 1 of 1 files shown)


jose-luis-rs:
Did you finish with this PR?

YanzhaoW:
> Did you finish with this PR?

Guess Igor has no problem with it. Then yes, please merge.

neuland/calibration/R3BNeulandCosmicTracker.cxx
```diff
@@ -67,12 +67,13 @@ namespace Neuland
...
-            : fFit("CosmicTracker:fFit", "pol1")
+            : fDistances{ 0. }
+            , fFit("CosmicTracker:fFit", "pol1")
         {
             fDistances.reserve(MaxNumberOfPlanes);
-            for (auto p = 0; p < MaxNumberOfPlanes; ++p)
+            while (fDistances.size() < MaxNumberOfPlanes)
             {
-                fDistances[p] = p * BarSize_Z;
+                fDistances.push_back(fDistances.back() + BarSize_Z);
```
klenze, line 76: I assume that CosmicTracker is instantiated only once per run. 

In that case, I prefer my variant (fDistances.resize(MaxNumberOfPlanes)) over your variant. 

I am aware that you are trying to avoid zero-initializing the vector only to overwrite the values again, but I find it less readable. (A larger fix would be to be to get rid of fDistances altogether.)

YanzhaoW: Do you mean the following code?
```cpp
CosmicTracker::CosmicTracker()
            : fFit("CosmicTracker:fFit", "pol1")
        {
            fDistances.reserve(MaxNumberOfPlanes);
            fDistances.resize(MaxNumberOfPlanes);
            for (auto p = 0; p < MaxNumberOfPlanes; ++p)
            {
                fDistances[p] = p * BarSize_Z;
```

This has pointer arithmetic, which is not very preferable according to the core guideline:
> Bounds.1: Don’t use pointer arithmetic

klenze: It seems your are correct in that the core guidelines also consider operator[] of (some) STL containers to be "pointer arithmetic". 

IMHO there is a world of a difference between std::vector::operator[] and adding offsets to raw pointers in that for the former I can simply enable boundary checks (which made this bug visible) and for the latter I can not. (Of course, this would then also affect line 470.)

klenze (approved):
Also, did clang-format force you to reformat stuff? e.g. around 422 or is there some subtle content change I am missing?

This is a bugfix, please merge without the customary week for reviews.

YanzhaoW:
> Also, did clang-format force you to reformat stuff? e.g. around 422 or is there some subtle content change I am missing?

Yeah, I guess so, but with clang-format 15.

jose-luis-rs:
Ready?

YanzhaoW:
@jose-luis-rs Yes, Please merge

PR #790, opened 2023-02-08
[merge-ready][hotfix]Handled R3BUcesbSource::Close() being called twice

klenze:


jose-luis-rs:
Thanks @klenze, is this modification affected by or related to other modifications in FairSoft/FairRoot?

klenze:
I think FairRoot started calling Close, while we also called Close in the destructor. So using any r3bsource with recent versions of FR would LOG(fatal) after the macro was otherwise completed.

YanzhaoW:
Great. Yeah, I just found the Close() is also called at FairRunOnline.cxx:351. Maybe we should also remove it in the destructor of R3BUcesbSource? 

I would also suggest adding some override keywords in the header file before merging to the branch.

klenze:
> Maybe we should also remove it in the destructor of R3BUcesbSource?

I think that destructors for anything which has a lifetime similar to the runtime of the program are generally overrated (excepting closing output files, perhaps). There is this cargo-cult programming pattern of "call delete on any member pointers during the destructor" and this can easily lead to double-frees. 

> I would also suggest adding some override keywords in the header file before merging to the branch.

I consider this patch to be a minimal hotfix which can be merged quickly so people get a working version of R3BRoot, simular to #788. While I agree that moving to using override everywhere is a good idea, I do not think that we should piggyback that change on a hotfix.

PR #759, opened 2023-01-19
[merge ready] Clang tidy: fixed some warnings, replaced C-style casts. 

klenze:
Should CI like this commit, please consider merging it.

PR #709, opened 2022-11-29
Califa ov assert

klenze:
Warns the people about #705. Thanks to @inkdot7.

PR #792, opened 2023-02-08
Moving califa from TClonesArray to STL, reworked R3BCalifaCrystalCal2Cluster.

klenze:
* New containers: std::map for R3BCalifaMappedData, R3BCalifaCalData. R3BCalifaClusterData is now a std::vector.
    * The new names are defined in FooData::default_container_name: CalifaMappedMap, CalifaCrystalCalMap, CalifaClusterVec
    * R3BCalifaLegacyWriter to create TCA copies of STL container items using the previously used names.
    * LLOG: location prefixing from R3BLOG without the weird syntax.
    * R3BCalifaGeometry is now a singleton. The first time it is initialized it needs the geometry number, afterwards
      Instance() can be fetched without arguments.
    * This means that you will have to call R3BCalifaGeometry::Instance(2021); in your macro to use clustering for experiment
      data instead of using cal2cluster->SetGeometryVersion.
    * Switched from TVector3 to XYZVector for most of califa, switched R3BCalifaDigitizer to also write to  std::map.
      (Code got shorter \o/)
    * Added some transient (not serialized) fields to R3BCalifaCrystalCalData. These are set by R3BCalifaCrystalCal2Cluster.
    * Cleaned up R3BCalifaCrystalCal2Cluster: Separated "are these two hits nearby?" into interface R3BCalifaClusterMatcher,
      provided 3 implementations. Please test. Split Exec into three methods:
      * FilterCrystals fills the candidate list from CrystalCalMap, removing duplicates from dual range and crystals below threshold
      * SortCandidates sorts the candidate list.
      * DoClustering finally does the clustering.
  * R3BCalifaCrystalCal2Cluster::SetClusterStartThreshold can be used to enforce a minimum energy for starting new clusters
      (idea from Gabriel).

Still needed:
   * Implement Gabriels randomization as a separate task.

This commit is currently on my branch, but I can rebase it to dev.


YanzhaoW:
I would suggest move the changes on .clang-format and .clang-tidy to separate PRs.

In which file for example do you use STL container to read and write data with root files? (sorry, there are too many changed files and I'm kinda disoriented trying to find the detailed implementation.)

klenze:
@YanzhaoW:

> I would suggest move the changes on .clang-format and .clang-tidy to separate PRs.

When I was starting to work on this, I had no idea when #736, #759 or #709 would get merged. I decided that rather working around bugs I already had fixed months ago I would rather use my own branch and let the devs cherry-pick any commits which might interest them. (Since then, I have seen some indication that #759 at least might be merged, so I will start working on rebasing that one next.)

The relevant changes of the PR  are in [this commit](https://github.com/R3BRootGroup/R3BRoot/pull/792/commits/2d31a0452f3ae1e0c5283f4e304d42e6c2934913). 

califa/calibration/R3BMapped2CrystalCal.* is probably a good example. 
R3BCalifaFooData::container_t is ``std::map<int, R3BCalifaFooData>`` for Foo in Mapped, CrystalCal.
The relevant parts are in Init. The input I get by:
```
    fCalifaMappedData = rootManager->InitObjectAs<decltype(fCalifaMappedData)>
           (R3BCalifaMappedData::default_container_name);
    assert(fCalifaMappedData);
```

The output is registered here:
```
    rootManager->RegisterAny(R3BCalifaCrystalCalData::default_container_name, fCalifaCryCalData, !fOnline);
```

I am happier about the RegisterAny method than about the InitObjectAs method. I would prefer the signature like ``template<typename T> void TDirectory::getObject(char* name, T*& ptr);``. Also, I don't think FairRoot should enforce constness on the pointer, but I fixed that by just marking the fields I want to change as mutable. (Also, InitObjectAs does not at all sound like Get. I am halfway confident that it will not create objects from thin air, but instead either fetch them from the vector of registered objects or search it in some input. Any template Get method is non-public so FR clearly does not want us to use them.)

YanzhaoW:
I actually didn't notice there are few new function templates added to FairRootManager. :D And really surprised that there were some efforts from FairRoot group trying to mitigate the type un-safety of TCA. 

I checked the implementation of function templates `InitObjectAs` and `RegisterAny`. Even though for now I could understand only a little bit, I don't feel very good about it, TBH. Passing null pointer is kinda NO GO in C++ and the only way it uses to ensure type safety (if I'm correct) is to pass typeid along the way. If so, it would be a very bad practice. But I'm not sure and need more time to play them around, get the details and see whether it's better than TCA.

klenze:
I was pointed to these methods because I made some complaints about the unsafety of TCAs. 

The original PR in FR is [here](https://github.com/FairRootGroup/FairRoot/pull/636). Seems like ALICE wanted that feature. 

The PR is from 2017, so it should probably not be judged to harshly against modern sentiments. 

(Personally, I think requiring a pointer to be passed by reference and then taking the address of that reference in RegisterAny seems somewhat fishy. But at least we have the typeid map and everything, which is better than the traditional YOLO approach seen so often with TClonesArrays.

PR #794, opened 2023-02-09
Hotfix: assertion in R3BUcesbSource::FillEventHeader fixed.

klenze:
@jose-luis-rs @hapol: This is a hotfix for an issue discovered by dynamic_cast which is affecting at least two collaboration members. Please consider merging it after a quick review once it has passed CI.


igasparic:
@klenze I think a similar fix is needed in r3bbase/R3BFileSource.cxx.

PR #816, opened 2023-02-19
[hot fix] adding dependencies for r3banalysis and evtivs in cmake

YanzhaoW:
Describe your proposal.

Error occurred during compilation. Fixed by adding few more dependencies in cmake
But why did our CI fail to catch this error?

---

Checklist:

### DIFF
--- a/analysis/CMakeLists.txt
+++ b/analysis/CMakeLists.txt
@@ -97,7 +97,7 @@ Set(LINKDEF AnaLinkDef.h)
 
 Set(DEPENDENCIES
 GeoBase ParBase MbsAPI Base FairTools R3BPassive R3BData Core Geom GenVector Physics Matrix MathCore R3BTraRene R3BTracking
-R3BSsd R3BLos R3BCalifa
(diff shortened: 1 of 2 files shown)


jose-luis-rs (commented):
Do you understand why we need R3BBunchedFiber and R3bMwpc here?

YanzhaoW:
> Do you understand why we need R3BBunchedFiber and R3bMwpc here?

https://github.com/R3BRootGroup/R3BRoot/blob/075013229f3d3a81c784346c5afef2e78392617d/analysis/online/R3BGeneralOnlineSpectra.cxx#L142-L143

https://github.com/R3BRootGroup/R3BRoot/blob/075013229f3d3a81c784346c5afef2e78392617d/analysis/online/R3BFibervsTofDOnlineSpectra.cxx#L73-L81

Both requires the two libraries you mentioned. Funny thing is when I compiled it in gsi server, which uses gcc, it doesn't show any error. But when I compiled it in my own laptop using clang (apple clang), compilation fails. One of reasons is gcc somehow optimise it away once it sees the pointers are always nullptr.

jose-luis-rs:
Ok, it is due the online tasks. Thanks.

jose-luis-rs (approved):

PR #822, opened 2023-02-23
GLAD: Added new setter SetFieldfromCurrent to scale the field

jose-luis-rs:
Added new setter SetFieldfromCurrent to scale the field according to the GLAD current, being the max. current 3583.81 A.

---

Checklist:

### DIFF
--- a/field/R3BGladFieldMap.cxx
+++ b/field/R3BGladFieldMap.cxx
@@ -24,6 +24,7 @@
 #include "TMath.h"
 
 #include "R3BGladFieldMap.h"
+#include "R3BLogger.h"
 
 using std::cerr;
 using std::cout;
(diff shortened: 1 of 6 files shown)


field/R3BGladFieldMap.cxx
```diff
@@ -74,39 +75,73 @@ R3BGladFieldMap::R3BGladFieldMap(const char* mapName, const char* fileType)
...
         fFileName += ".dat";
+    }
 
     fType = 2;
 }
 
+void R3BGladFieldMap::SetFieldfromCurrent(double current)
+{
+    R3BLOG_IF(fatal, TMath::Abs(current) > 3583.81, "GLAD current cannot be larger than 3583.81 A.");
+    fScale = current / 3583.81;
```
morze00, line 112: fScale should be a negative number

jose-luis-rs: Yes, the current must be negative. For FAIR it will be negative and positive depending on the configuration. 
What is the right convention?
I am assuming a positive current for the standard FAIR configuration and negative one for the current setup.

hapol: Is this ok or should we change the sign?

jose-luis-rs: For the root macros one has to call the R3BGladFieldMap class with these lines:

R3BGladFieldMap *magField = new R3BGladFieldMap("R3BGladMap");
magField->SetFieldfromCurrent(fMeasCurrent);

where fMeasCurrent corresponds to the measured GLAD current. For the Phase-0 setups this must be negative and in Amperes.

PR #806, opened 2023-02-14
[WIP] Migrate to GitHub-hosted runners

kresan:
Describe your proposal.

- Run R3BRoot CI on the GitHub-hosted Ubuntu 18.04
- Use FairSoft and FairRoot from CVMFS (versions will be updated soon)
- Only cvmfs package needs to be installed and configured - takes 1 min
- 2 cores are available per runner

Mention any issue this PR is resolved or is related to.

---

Checklist:

* [x] Rebased against `dev` branch
* [x] My name is in the resp. CONTRIBUTORS/AUTHORS file
* [x] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

PR #825, opened 2023-02-28
CI/CD: Add clang-tidy github auto review

YanzhaoW:
Add clang-tidy auto review for all pull requests.

Version info:
* clang: 15
* FairRoot: 18.8.0
* FairSoft: nov22

Checking filetype: '*\.(cpp|cc|c\+\+|cxx|cl|h|hh|hpp|m|mm|inc)'

ROOT MACROS with .C will not be checked with clang-tidy.

@klenze We can open another issue to discuss what should be put inside .clang-tidy.

check "Files changed" tab above to see the effect.

~Task failure is intentional because of the new bad code added with the PR.~

Current docker image contains FairSoft/FairRoot/Clang15 and has size of ~8GB in total. Therefore the clang-tidy checking needs to take a while (~3 mins ). I will try to create another docker image using cvmfs of FairSoft. ( FairRoot still needs to be compiled and built in the container since its version in cvmfs is bit old. )

---

Checklist:

### DIFF
--- a/.github/workflows/clangtidy.yml
+++ b/.github/workflows/clangtidy.yml
@@ -13,26 +13,40 @@ jobs:
   clang-tidy:
     runs-on: ubuntu-latest
     container: 
-      image: yanzhaowang/fairroot:v18.8.0
-      options: --user root
+      image: yanzhaowang/cvmfs_clang:v15
+      volumes:
+        - /tmp:/cvmfs
+      env:
+        CVMDIR: /cvmfs/fairsoft.gsi.de
+      options: --user root --privileged  --ulimit nofile=10000:10000 --cap-add SYS_ADMIN --device /dev/fuse
     steps:
-    - name: install deps
-      run: |
-        apt-get update -y && apt-get install -y git
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
+    - name: mount cvmfs
+      run: |
+        wget https://cernbox.cern.ch/remote.php/dav/public-files/RmnTeeOZpYjCJQb/masterkey.gsi.de.pub
+        mv masterkey.gsi.de.pub /etc/cvmfs/keys
+        cp .githubfiles/fairsoft.gsi.de.conf /etc/cvmfs/config.d
+        mkdir -p ${CVMDIR}
+        mount -t cvmfs fairsoft.gsi.de ${CVMDIR}
+
     - name: cmake configure
       run: |
+        export SIMPATH=${CVMDIR}/debian10/fairsoft/nov22p1
+        export FAIRROOTPATH=${CVMDIR}/debian10/fairroot/v18.8.0_fs_nov22p1
+        export PATH=${SIMPATH}/bin:$PATH
         git config --global --add safe.directory $GITHUB_WORKSPACE
         git clone https://github.com/R3BRootGroup/macros.git
-        cmake . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=on 
+        cmake . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=on -DBUILD_GEOMETRY=OFF -DUSE_DIFFERENT_COMPILER=TRUE
 
     - name: clang-tidy check
       run: |
         mkdir -p clang-tidy-result
-        git diff -U0 HEAD^ | clang-tidy-diff-15.py -p1 -path build -export-fixes clang-tidy-result/fixes.yml |\
+        git diff -U0 HEAD^ | clang-tidy-diff-15.py -p1 -path build \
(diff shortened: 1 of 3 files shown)


kresan (approved):

klenze:
``modernize-use-trailing-return-type``
I am actually opposed to that one. I think trailing return types have their uses (such as when the return type depends on auto parameters of the function), but I don't think they generally increase readability. Also, my guess is that perhaps one in three of our devs can use them. 

In general, we should probably define some "best practices" at some point. I personally did not know about std::accumulate when I encountered it in your code, but I could figure out what it does quickly enough from the docs. For people not experienced with higher order functions and lambda expressions, this will be a major roadblock for understanding the code. 
(Naturally, my self-filling histograms are full of code which is terribly unreadable, but this is mostly because I try to provide a fast and simple syntax for that. If it does not work for someone, they can easily go back to spending a few kilobytes to declare arrays of histogram pointers. By contrast, in my work on califa clustering, I have gotten rid of a lambda in favor of a function so the code gets more readable.)

We are not a C++ shop where every contributor can be expected to stay up to date on the latest trends in C++. Personally, I think that we should focus on bugprone behavior first. Trailing return types as a default are more a matter of taste. 

Anyhow, good work implementing clang-tidy in CI. For format, we had the policy "whoever touches a file is responsible for making that file pass CI, even if it would not have passed before". This policy was not good for clang-format, and it will be disastrous for clang-tidy (where most complaints can't be fixed automatically). 

But on the conditions that
* the check is bugprone rather than a matter of taste
* it is something which we can realistically teach our contributors to avoid (e.g. pointers as arrays, but not raw pointers when interfacing root)
* the check is fixed in the codebase

I am generally in favor of adding more checks. One could always run a variety of checks over the codebase and pick whatever looks most painful. 

Also, clang-tidy probably will not check macros. Which is yet another great reason to switch to compiled code.

YanzhaoW:
@kresan I switched into cvmfs-based container but the speed is actually not increased very much. 

Yes, with fully installed image, downloading take around 2 mins and cmake configuration takes 8s. In total it takes 3m 18s. With cvmfs mounted container, downloading takes 32 s but cmake takes 1 min 8 s. In total it takes 2m 32s. Things could be worse for the cvmfs based container if compilation and testing are involved.

Therefore I'm not so sure mounting FairSoft&FairRoot using cvmfs is a good idea to boost test speed. It works great on HPC cluster due to the high speed connection. But on github hosted servers, there could be no advantage at all.

YanzhaoW:
Hi, @jose-luis-rs. It's done. Let's launch it.

 Could you just merge with two commits. Both of them are very important: one is using the fully installed image; another (current) is using the cvmfs mounted images. If second one doesn't work in the future, I can easily switch to the first.

jose-luis-rs:
In this case I don't see any problem. But you are still doing modifications...

YanzhaoW:
> In this case I don't see any problem. But you are still doing modifications...

I'm very sorry. I just know I can turn off the the geo build. Now it's truly done. ;D

inkdot7:
@YanzhaoW  Just a side-line note about the CI build speed: I think a few minutes is just fine.  Once things are changed to use github cloud builders (#806), all CI jobs should be run in parallel, such that the visible time is just the time of the longest job.  With a project of this size some 5 (to 10) min total CI time would be very decent.

inkdot7:
Btw, what would it be that limits the cvmfs speed?

I guess it is not first downloading a multi-GB compressed image, but that the issue rather is latency of long network distances as it access things serially file-by-file?

This one could try by timing the cvmfs mount vs running of the programs twice.

If the large time is the first running of the program, one could come up with a 'seed-file': find out which are the files typically loaded from cvmfs during typical CI jobs.  I suppose strace could help with that.  Then after mounting the cvmfs one could issue some wide parallel `cat filelist | xargs -n10 -P100 cat > /dev/null` to force the files loaded quickly .

(I'm not suggesting to delay merging this PR to find these things out.  Rather a optimisation that can be done afterwards.)

YanzhaoW:
@inkdot7 Yes, few minutes should be fine. And if we switch to Github hosted runner, different tasks can be run in parallel. If I understand you correctly, I'm not so sure loading the specific files in the beginning really helps. If cvmfs doesn't remove the file caches until the end of the program and if the necessary files are only loaded once, putting the loading time into the compilation time or moving it to the beginning of the CI probably doesn't make any difference in total. And also in this case, using a fully installed docker image won't be much slower. I will make some tests about the speed of each method.

YanzhaoW:
@jose-luis-rs ready to be merged

jose-luis-rs:
@YanzhaoW , please, solve the conflicts

YanzhaoW:
@jose-luis-rs Yes, I have done that.

PR #829, opened 2023-03-03
Added macros to generate the NeuLAND ROOT geometry with 26 planes

jose-luis-rs:
New ROOT geometry provided by Igor with 26 planes for NeuLAND

---

Checklist:


neuland/geobase/create_neuland_geo.C
```diff
@@ -0,0 +1,76 @@
...
+            else
+            {
+                volNeuland->AddNode(volPaddle, nindex, new TGeoCombiTrans(a, 0, b, rot90));
+            }
+        }
+    }
+
+    TGeoVolume* top = new TGeoVolumeAssembly("TOP");
+    gGeoManager->SetTopVolume(top);
+    top->AddNode(volNeuland, 1, new TGeoCombiTrans(0., 0., distance + nPlanes + gPaddleDistance, rot0));
```
YanzhaoW: No, distance should not be set here. The value of distance between the detector and source should be left to user to decide.

YanzhaoW:
Is there any cmake cache variable I can set to disable auto generation of the geometry file?

neuland/geobase/create_neuland_geo.C
jose-luis-rs: Yes, we do this with in-beam detectors because they could be moved during the beam time. But this is not the case for NeuLAND, right? Did you move it during the last 4-6 experiments?

From my side there is not problem with setting its position in (0,0,0), but in this case all the users must have very clear the position of this detector to avoid misunderstandings. Maybe NeuLAND experts (@igasparic) can comment on what they want to have here.

jose-luis-rs:
Yes, it is already implemented in the dev branch -> #778

neuland/geobase/create_neuland_geo.C
YanzhaoW: Even if we will never move it, it's still kind of bad to define the distance value here, for the following reasons:
1. Distance value can also be set through R3BNeuland constructor. One quantity should only be defined through one method. If the value can be set here and there, it's bound to cause trouble in the future.
2. The geometry file can also be used in the simulation, which could use different distances.
3. Geometry file should only contains geometrical information. We don't need to create different geo files for the exactly same geo configurations of Neuland with different distances., like neuland_geo_13dp_1600cm, neuland_geo_13dp_1500cm, neuland_geo_13dp_1400cm ...

jose-luis-rs: Yes, for me it is also better to have something general for all the detectors and passive elements like GLAD.

igasparic: I agree with Yanzhao. I would leave distance to a user to decide. For instance when a new double plane is added then we usually move other planes as well to pack them tightly. So although the holding frame is fixed the detector position with respect to the cave changes.

jose-luis-rs:
In the standard root geometry the position (0,0,0) is in the center of the detector as shown in the attached figures.

Screenshot from 2023-03-04 15-11-41
Screenshot from 2023-03-04 15-12-09

So, for the use of this root geometry in simulations, the user must define the NeuLAND position in the root macro like this:

run->AddModule(new R3BNeuland(fNeuLandGeo, {0., 0., fDistance}));

where fDistance is the distance with respect to the nominal target position.

neuland/geobase/create_neuland_geo.C
jose-luis-rs: Done!

jose-luis-rs: Please, add to the R3B-Wiki the position of NeuLAND for each experiment, so users can take the right values for simulations, analysis, etc.

neuland/geobase/create_neuland_geo.C
```diff
@@ -0,0 +1,75 @@
...
```
YanzhaoW: Wait, what's the meaning of this:
```cpp
new TGeoCombiTrans(0., 0., distance + nPlanes + gPaddleDistance, rot0)
```
Especially the summation `distance + nPlanes + gPaddleDistance`? Currently to generate Neuland geo file, we use `/macros/r3b/geo/create_neuland_v3_geo.C`. Is this MACRO file based on that?

neuland/geobase/create_neuland_geo.C
YanzhaoW: I would say remove the distance variable in the MACRO. If someone in the future see that he can define a distance in the macro, he will either probably define the same distance value again or be puzzled why there are two ways to set the distance value.

neuland/geobase/create_neuland_geo.C
jose-luis-rs: I don't know about that, but you have the figures above to see the reference point (0,0,0)

YanzhaoW: Adding the number of planes with a distance value looks really strange to me. It makes sense if it's `nPlanes*gPaddleDistance`.

jose-luis-rs: Yes, I agree. Maybe @igasparic could comment on this.

igasparic: It looks this '+' is a typo which was not corrected in all geometry versions. It should be nPlanes * gPaddleDistance. I think this went unnoticed because in a simulation macro the distance is usually overwritten anyhow.

neuland/geobase/create_neuland_geo.C
jose-luis-rs: Done!

neuland/geobase/create_neuland_geo.C
jose-luis-rs: I have removed it since the user will decide the position.

PR #830, opened 2023-03-07
Adding a factory function for ROOT owning objects

YanzhaoW:
## Factory function
Explicitly specifying the ownership of the ROOT objects (TH1 and TF1).

### Usage:
Instead of 
```cpp
auto* hist = new TH1D("hist", "hist", 100, 0., 10.); // warning from Clang-tidy
```
use
```cpp
auto* hist = r3b::root_owned<TH1D>("hist", "hist", 100, 0., 10.); // OK
```

Feel free to comment on the implementation (including naming) @klenze 

Some info about the ownership in ROOT : [object ownership](https://root.cern.ch/root/htmldoc/guides/users-guide/ObjectOwnership.html)

## Change in r3bbase/CMakeLists.txt
Explicitly writing out the source file names, according to cmake document:
> We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

---

Checklist:

### DIFF
--- a/r3bbase/CMakeLists.txt
+++ b/r3bbase/CMakeLists.txt
@@ -37,10 +37,34 @@ file(GLOB SRCS
 *.cxx 
 )
 
-# fill list of header files from list of source files
-# by exchanging the file extension
-CHANGE_FILE_EXTENSION(*.cxx *.h HEADERS "${SRCS}")
-install(FILES "R3BLogger.h" DESTINATION include)
+set(SRCS
+    R3BCoarseTimeStitch.cxx
+    R3BDataPropagator.cxx
+    R3BDetector.cxx
+    R3BEventHeader.cxx
+    R3BEventHeaderPropagator.cxx
+    R3BFileSource.cxx
+    R3BLogger.cxx
+    R3BModule.cxx
+    R3BTcutPar.cxx
+    R3BTsplinePar.cxx
+    R3BWhiterabbitPropagator.cxx
+    )
+
+set(HEADERS 
+    R3BCoarseTimeStitch.h
+    R3BDataPropagator.h
+    R3BDetector.h
+    R3BEventHeader.h
+    R3BEventHeaderPropagator.h
+    R3BFileSource.h
+    R3BLogger.h
+    R3BModule.h
+    R3BShared.h
+    R3BTcutPar.h
+    R3BTsplinePar.h
+    R3BWhiterabbitPropagator.h
+    )
 
 Set(LINKDEF BaseLinkDef.h)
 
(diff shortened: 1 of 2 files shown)


YanzhaoW:
@jose-luis-rs Ready to be merged.

jose-luis-rs:
Thanks @YanzhaoW, could you add more objects like TCanvas, TH2, TGraph, etc ?

YanzhaoW:
Hi @jose-luis-rs TH2 is already included since it inherits from TH1. I'm not sure TGraph and TCanvas are also auto deleted by ROOT. Did you find some sources about this? I may do some tests later to see whether they are owned by other root objects.

jose-luis-rs:
Yes, you are right with the TH2, I forgot it! Sorry!

TGraph is like TH1&TF1, TCanvas I am not sure, maybe the right class is TPad. 

Please, don't forget the usual license header for new files.

klenze:
From the ROOT object ownership web page, I don't see anything on TGraphs. Also, I would be somewhat amazed if TGraphs (which are not be required to have names) would be stored in TDirectories by default. 

In general, TH1 can work with unique_ptr, provided that the lifetime of *gDirectory (at histogram creation time) is longer than that of the histogram. 

For new code, I would prefer not storing the histogram in gDirectory at all:
```
int Foo::Init() // or constructor
{
  auto oldGDirectory=gDirectory;
  gDirectory=nullptr;
  this->fH=std::make_unique<TH1I>("foo", "", 100, 0, 1);
  gDirectory=oldGDirectory;
}
```
(Alternatively, store the previous value of TH1::AddDirectoryStatus() and temporarily overwrite TH1::AddDirectory). 

In the long run, I feel histograms should be handled quite differently. If I have five quantities, each with their own axis range, and want to plot each of them on their own as well as all pairs of them, that are already 15 different axis combinations and it is up to the user to ensure that each Fill statement (likely 1000 lines further down) fills exactly the right quantities. And that is even before we get to the fact that many histograms are per channel, where channel is some n-tuple of integers.

(My own implementation works by using storing everything there is to know about an axis in a type so that the maximum amount of information is already present at compile time. It would be possible to build auto-fillable histograms without many templates (e.g. ``struct Axis{ Axis(std::function<...>, int bins, double min, double max) ... };``. This will probably have a worse performance, but on the other hand template mistakes lead to very unreadable error messages.)

From TPad.cxx, TPad only seems to interact with gDirectory in TPad::Print().

Of course, the ownership of FairRoot stuff is also not straightforward. 

> Explicitly writing out the source file names, according to cmake document

Agreed. From a usability point of view, the ability to quickly disable a class which won't compile without actually removing  it is desirable.

YanzhaoW:
I checked TCanvas and TPad. They are ok to be deleted by the user. TGraph seems also not owned by ROOT as well.

jose-luis-rs:
Ok, thanks @YanzhaoW.

YanzhaoW:
@jose-luis-rs I have added the license. Please merge it to dev branch.

PR #834, opened 2023-03-28
Tests for CIs

jose-luis-rs:


jose-luis-rs:
It seems to work correctly, although I was expecting 5 red lights!

PR #835, opened 2023-03-29
New clang-format CI independent of repository

jose-luis-rs:
---

Checklist:

### DIFF
--- a/util/clang-format-check.sh
+++ b/util/clang-format-check.sh
@@ -4,6 +4,7 @@ test "$1" == "--ci" && AUTOFIX=1 && CI=1 && shift
 
 CLANG_FORMAT=${1:-clang-format}
 
+# Starting
 echo -n "Running clang-format checks, version: "
 ${CLANG_FORMAT} --version
 
@@ -12,14 +13,9 @@ if [ 0 != $? ]; then
     exit 1
 fi
 
-if [ "$TRAVIS" != "true" ] ; then
-  # Not in a pull request, so compare against parent commit
-  base_commit="origin/dev"
-  echo "Checking against parent commit $(git rev-parse $base_commit)"
-else
-  base_commit="$TRAVIS_COMMIT_RANGE"
-  echo "Checking against commit $base_commit"
-fi
+git fetch --all
+base_commit="origin/dev"
+echo "Checking against parent commit $(git rev-parse $base_commit)"
 
 # To simplify CI debugging, tell what files are considered.
 echo "--- Listing all changed files:"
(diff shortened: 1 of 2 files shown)


util/clang-format-check.sh
```diff
@@ -1,9 +1,13 @@
 #! /bin/bash
 test "$1" == "--autofix" && AUTOFIX=1 && shift
-test "$1" == "--ci" && AUTOFIX=1 && CI=1 && shift
+#test "$1" == "--ci" && AUTOFIX=1 && CI=1 && shift
 
 CLANG_FORMAT=${1:-clang-format}
 
+# Install clang-format-15 for the check
+sudo apt-get install clang-format-15
```
YanzhaoW: Should this be in main.yml?

jose-luis-rs: Yes, I guess I can move it.

YanzhaoW: Thanks. I saw from Github history that Philip (@klenze ) wrote the original shell script. Maybe you could wait for his feedback?

jose-luis-rs: I though that the script was created by Jan Mayer or Dima Kresan, but @klenze did modifications recently introducing the error. Basically the script didn't detect the wrong files.

The new version solves the problem.

YanzhaoW:
@jose-luis-rs Could you put the clang-format checking under the workflow "static analysis"? Maybe you should also change the filename from "clangtidy.yml" to "static_analysis.yml".

util/clang-format-check.sh
klenze: What I was trying to do with the --ci flag was making clang-format run with -i (just like for --auto-fix), then upload the resulting patch, so people without a suitable clang-format could just download and apply it. 

From my understanding, the main problem with the previous code seems to be in line 31,
```
filesToCheck="$(git diff --name-only ${base_commit} | grep -e '.(\.C\|\.cpp\|\.cxx\|\.h)$' || true)"
```
which does not work as intended because in basic regular expressions '(' and ')' are not special unless prefixed by backslash. I am doubtful that I wrote that line. 

Personally, I would use extended RE: ``grep -E '[.](cxx|h|C)$'`` seems to work for me. 

(It is also possible that for some weird reason, $fail is set while git status returns the empty string. If that was the case then my version would indeed seem to accept wrong formats with --ci. In that case, replacing the if condition in line 51 would have been sufficient.)

jose-luis-rs: Ok, thanks @klenze. I will check it next week after holidays.

jose-luis-rs:
> @jose-luis-rs Could you put the clang-format checking under the workflow "static analysis"? Maybe you should also change the filename from "clangtidy.yml" to "static_analysis.yml".

Done

jose-luis-rs:
Ready!

PR #836, opened 2023-03-30
Changed Y position calculation for mwpc1

ManuelXarepe:
We noticed that for the S522/S509 experiments, the Y position of MWPC1 seems to be set to -1000 mm in almost 50% of all data.

After looking at the code, we noticed that when the function GetPositionY is called it requires that at least 3 pads have a charge different than zero which is not the case most of the time.

The implemented code instead of calling the GetPositionY function, calls a new function named GetPositionYS522 which calculates the center of mass for the existing pads. These changes have only been applied to Y of mwpc1 as is the most problematic. I believe these changes should not affect anyone else as it is locked to expID S522 or S509.

Checklist:

### DIFF
--- a/mwpc/mwpc1/R3BMwpc1Cal2Hit.cxx
+++ b/mwpc/mwpc1/R3BMwpc1Cal2Hit.cxx
@@ -34,6 +34,9 @@
 #include "R3BMwpcCalData.h"
 #include "R3BMwpcHitData.h"
 
+constexpr int S522 = 522;
+constexpr int S509 = 509;
+
 // R3BMwpc1Cal2Hit: Default Constructor --------------------------
 R3BMwpc1Cal2Hit::R3BMwpc1Cal2Hit()
     : R3BMwpc1Cal2Hit("R3BMwpc1Cal2Hit", 1)
@@ -107,6 +110,17 @@ Double_t R3BMwpc1Cal2Hit::GetPositionX(Double_t qmax, Int_t padmax, Double_t qle
                                                     (2 * TMath::SinH(TMath::Pi() * fwx / a3)));
 
     return (-1. * padmax * fwx + (fSize / 2) - (fwx / 2) - a2); // Left is positive and right negative
+}
+
+/* ----   Protected method to obtain the position Y ---- */
+Double_t R3BMwpc1Cal2Hit::GetPositionYCoG(Double_t qmax, Int_t padmax, Double_t qdown, Double_t qup) const
+{
+   const double pos1 = (padmax-1) * fwy - (fSize / 2) + (fwy / 2);
+   const double pos2 = padmax * fwy - (fSize / 2) + (fwy / 2);
+   const double pos3 = (padmax+1) * fwy - (fSize / 2) + (fwy / 2);
+   const double center_of_mass = (pos1*qdown + pos2*qmax + pos3*qup)/(qmax+qdown+qup);
+
+   return center_of_mass;
 }
 
 /* ----   Protected method to obtain the position Y ---- */
@@ -203,8 +217,12 @@ void R3BMwpc1Cal2Hit::S467()
         // Obtain position Y ----
         qdown = fy[padmy - 1];
         qup = fy[padmy + 1];
-        if (qdown > 0 && qup > 0)
+        // TODO: check if systematic statistics loss is experiment specific, put general solution if not
+        if (header->GetExpId()==S522 ||  header->GetExpId()==S509){
+            y = GetPositionYCoG(qmy, padmy, qdown, qup);
+        } else if (qdown > 0 && qup > 0){
             y = GetPositionY(qmy, padmy, qdown, qup);
+        }
 
         AddHitData(x, y);
     }
--- a/mwpc/mwpc1/R3BMwpc1Cal2Hit.h
+++ b/mwpc/mwpc1/R3BMwpc1Cal2Hit.h
@@ -95,6 +95,8 @@ class R3BMwpc1Cal2Hit : public FairTask
     Double_t GetPositionX(Double_t qmax, Int_t padmax, Double_t qleft, Double_t qright);
     /** Private method to obtain the position Y **/
     Double_t GetPositionY(Double_t qmax, Int_t padmax, Double_t qdown, Double_t qup);
+    /** Private method to obtain the position Y **/
+    [[nodiscard]] Double_t GetPositionYCoG(Double_t qmax, Int_t padmax, Double_t qdown, Double_t qup) const ;
 
   public:
     // Class definition


mwpc/mwpc1/R3BMwpc1Cal2Hit.h
```diff
@@ -28,6 +28,8 @@
 
 #define Mw1PadsX 64
 #define Mw1PadsY 40
+#define S522 522
+#define S509 509
```
YanzhaoW: Using C macro is a bad idea. If you just want to define some int variables, please do like this:
```cpp
constexpr int Mw1PadsX = 64;
```
And place them in the source file (.cxx). But of course if you really need them in the header file, you could also place them over there.

mwpc/mwpc1/R3BMwpc1Cal2Hit.cxx
```diff
@@ -203,8 +214,12 @@ void R3BMwpc1Cal2Hit::S467()
...
         qdown = fy[padmy - 1];
         qup = fy[padmy + 1];
-        if (qdown > 0 && qup > 0)
-            y = GetPositionY(qmy, padmy, qdown, qup);
+        if (header->GetExpId()==S522 ||  header->GetExpId()==S509){
+         y = GetPositionYS522(qmy, padmy, qdown, qup);
+	    } else
+        if (qdown > 0 && qup > 0){
+         y = GetPositionY(qmy, padmy, qdown, qup);
+	    }
```
YanzhaoW: Here are some indentation problems, which are also suggested by clang-tidy. I'm kind of surprised that our clang-format checking doesn't work here (@klenze @jose-luis-rs).

Complexity level warning in this case will be fixed by my new PR #837.

jose-luis-rs: Yes, the clang-format problem is solved in #835

ManuelXarepe: Indentation problems have been fixed.

mwpc/mwpc1/R3BMwpc1Cal2Hit.h
ManuelXarepe: There is no reason for this to be in the header file, has been moved to the source file.

YanzhaoW:
@jose-luis-rs Could you rerun the static analysis workflow?

jose-luis-rs:
> @jose-luis-rs Could you rerun the static analysis workflow?

Hi, @ManuelXarepe must rebase the mainrepo to update the files, otherwise it will fail again.

klenze:
@ManuelXarepe: Do you know for a fact that there are other experiments working with R3BRoot dev which depend on the previous behavior? If not, consider replacing it everywhere. 

Also, I struggle a bit to understand what the original code was trying to do. A short comment why the use of inverse hyperbolic trigonometric functions is essential to accomplish the goal (or just ``// see Thesis / TDR, page X``) would have been appreciated. 

From my understanding, the previous code calculated the position based only on padmax if at least one of qleft or qright are zero and summons Elder Gods from the Void Behind The Stars otherwise. (Of course, the GetPositionX will still happily divide through zero, mixing ieee 754 infinities into all these hyperbolic functions.) At least the "if qleft is 0, we also disregard qright" does not seem sensible to me. 

Perhaps try to contact the original author (git blame) and ask them what they were trying to accomplish. If the answer is not convincing, replace both GetPositionX and GetPositionY with your arithmetic mean implementation. 

Also, I am assuming that you wrote S522 to get rid of the magic constant warning. IMHO, it would also be acceptable to mute that warning by using //NOLINT after the if statement. 

In general, I advise some reluctance to add experiment specific code. Either the ATanH stuff is terribly important. In that case one would have to figure out how to fix it. Or it is not important. In that case, replacing it with simpler code should happen everywhere. 

Edit: I think that the ``if (qdown > 0 && qup > 0)`` in line 222 is causing most of your original grief. That if should not be required, as the previous GetPositionY handles the case of qdown or qup being zero. (I am assuming that qfoo can not get negative, here.) "Set y just based on padmax" does seem more reasonable than "Keep whatever value for y was previously calculated". (The same problem exists for x.)

(Also, this is why you should declare your variables in the smallest scope you can. If you declare (and invalid-initialize (though -1000 makes a poor invalid value for a double, try NAN)) a variable you only need inside the loop outside a loop, this opens up the possibility that instead of using the invalid value, you end up using the value it was assigned to in a previous loop iteration. (Also, the github editor needs parenthesis matching.))

Summary of my response:
* I understand your fix, it seems fine (but should probably be applied everywhere, and also to X)
* I think the original code is buggy, error-prone and should not have been merged into dev. :-)

ManuelXarepe:
@klenze Thank you for the feedback. I address the comments you raised bellow.

> @ManuelXarepe: Do you know for a fact that there are other experiments working with R3BRoot dev which depend on the previous behavior? If not, consider replacing it everywhere.

Even though I do not know if other experiments are using the source code here discussed, I would prefer to not change the existing code without knowing for sure it will not affect previous experiments.

> Also, I struggle a bit to understand what the original code was trying to do. A short comment why the use of inverse hyperbolic trigonometric functions is essential to accomplish the goal (or just `// see Thesis / TDR, page X`) would have been appreciated.
> 
> From my understanding, the previous code calculated the position based only on padmax if at least one of qleft or qright are zero and summons Elder Gods from the Void Behind The Stars otherwise. (Of course, the GetPositionX will still happily divide through zero, mixing ieee 754 infinities into all these hyperbolic functions.) At least the "if qleft is 0, we also disregard qright" does not seem sensible to me.

The reason why this function is used can be found here, https://www.sciencedirect.com/science/article/abs/pii/0168900295006044?via%3Dihub, which I believe worked for previous experiments, but in our case, 50 % of the time only 2 pads have charge values, falling the condition that calculates the position and sets it to the default value of -1000. Because this mostly only affected the Y position I only corrected it there, as this improvement worsen a bit the resolution, which I'm comfortable with when it means an increase of 50 % in statistics, but not so much for less than 2 %.

> In general, I advise some reluctance to add experiment specific code. Either the ATanH stuff is terribly important. In that case one would have to figure out how to fix it. Or it is not important. In that case, replacing it with simpler code should happen everywhere.

I tend to agree that having experiment-specific code is not beneficial to the code base, however, in this case, the main goal was to have a fix that would not affect anyone else, and this was the only way I knew of achieving this.

Regarding the rest of your comments, I do not have enough experience in coding to discuss how bad or good the previous code is. 
However, I'm willing to open an issue presenting this problem and ask for an improvement, which I would be more than happy to help fix, after I finish the code development and improvements for the RPC detector.

With that being said, do you have any objections to this PR being merged as is?

klenze:
Thanks for the link. 

How about this:
* If all of the q's are nonzero, keep using the SECHS method from the paper. 
* If at least one of them is zero, use the center of gravity method (COG), which is what you implemented. 

That would seem to be a strict improvement of the status quo (where SECHS was used for PosY for q!=0 and y was a garbage value otherwise), while also fixing your problem. 

An even better way would be to calculate a2 correctly for the limit of qdown->0. However, chatGPT was not able to find a consistent limit for that case for me. (it claimed -> infinity for the original formula, and -> 0 for a simplification. I would have expected some simple function of qmax, qup and fwy). 

Or you could set qdown=max(1e-10*qmax, qdown). That way, you would not need to switch models or calculate any limits. (If you do, please test the outputs to see if they are reasonable). 

All in all, I do not have any strong objection to this PR being merged as it is, but I still feel it would be better to fix the behavior once and for all instead of inserting a workaround for specific experiments and let other experiments run into that issue again. 

Edit: Merging this PR, perhaps with a //TODO: and then later rewriting the code more thoroughly would be fine with me.

ManuelXarepe:
@klenze I added the TODO message.

jose-luis-rs:
Hi, sorry I was traveling and just saw the messages now.

Originally the code was written to track medium-mass and heavy fragments where we always have 3 active pads and this was done in 2019 for the fission experiments. Of course, in 2019 we didn't know that after our fission experiments the collaboration would use these detectors for other experiments, in particular for light ions where the current conditions could fail.

For the fission experiments we need 3 active pads to get good position resolutions ~250um (FWHM) and initially these detectors are "only" used for the ionization chamber calibrations (positions and angle). After these calibrations, the MWPCs are not required for analysis.

>How about this:
>  -If all of the q's are nonzero, keep using the SECHS method from the paper.
>  -If at least one of them is zero, use the center of gravity method (COG), which is what you implemented.

This depends on the experiment, in terms of resolutions it is a bad idea. Why? because when we have:
- 3 active pads ~250um
- 2 active pads ~ 1.5mm (maybe a bit better!)
- 1 active pad -> 3.125mm

According to your proposal, we could mix positions with very different resolutions that will introduce deviations in the calibration of other detectors. 

For the analysis of S509 and S522 I think that the function introduced by Manuel makes what they need for the analysis, so go ahead. Just a basic suggestion @ManuelXarepe to make your new function more general:
> GetPositionYS522(...)  -> GetPositionY_COG(...)  
(or similar because this function can be used in many experiments with light ions and I think that it is better to indicate the name of the used algorithm)

> Edit: Merging this PR, perhaps with a //TODO: and then later rewriting the code more thoroughly would be fine with me.

Yes, I agree because for FAIR we will need a new implementation ...

klenze:
Before the PR, the algorithm for posY seems to be:
* For 3 active pads, use SECHS
* ~~Otherwise, reuse the previous y value (which may or may not be the initial y=-1000)~~
* Otherwise, set it to -1000 
(Edit: I assumed the optional assignment of y was in a loop, but it is just in an if. So this works if we don't need multihit capabilities.)

The resolution of for single hits in the y direction (pitch 5mm) is probably sigma=5mm / sqrt(12)=1.44mm. Which I agree is likely much worse than what one would get with three pads. 

One could just store the number of pads used for position reconstruction in X/Y direction with the hit. Then other detectors using the MWPC can decide which hits they want to consider. Or we could have an 

```
enum PosMethod
{
    fBest,     // pick best method for every hit
    fCoGOnly,  // always CoG
    fSECHSOnly // set pos to NAN for less than 3 pads
} fPosMethod;
```
(There are probably more modern ways to write enums.)

Also, there is some code duplication wrt x, y. 
I would suggest
```
double pos_adjust_cog(double qmax, double qneg, double qpos) 
// in units of pitch, relative to middle strip
{
    return -qneg+qpos/(qmax+qneg+qpos);
}

double pos_adjust_sechs(double qmax, double qneg, double qpos)
{
   // basically, return a2 for fwy=1. (should be linear in fwy)
}

double pos_adjust(...)
{
   if (fPosMethod!=fCoGOnly && qpos>0 && qneg > 0)
      return pos_adjust_sechs(...);
   if (fPosMethod==fSECHSOnly)
      return NAN;
   return pos_adjust_cog(...);
}

double GetPosY(double qmax, int padmax, double qdown, double qup)
{
   return padmax * fwy - (fSize / 2) + (fwy / 2) + fwy*pos_adjust(qmax, qdown, qup);
}
double GetPosX(double qmax, int padmax, double qleft, double qright)
{
   return ...;
}

```

Also, I guess that S467() should be renamed so that @ManuelXarepe does not need to lie about the experiment he is analyzing, fX & friends should be local variables (std::array or std::vector), the array calData should not exist at all (an in-loop pointer to the current element replacing calData[i] is sufficient). Also ES.5. :-)

jose-luis-rs:
> Before the PR, the algorithm for posY seems to be:
> 
>     * For 3 active pads, use SECHS
> 
>     * ~Otherwise, reuse the previous y value (which may or may not be the initial y=-1000)~
> 
>     * Otherwise, set it to -1000
>       (Edit: I assumed the optional assignment of y was in a loop, but it is just in an if. So this works if we don't need multihit capabilities.)
> 
> 
> The resolution of for single hits in the y direction (pitch 5mm) is probably sigma=5mm / sqrt(12)=1.44mm. Which I agree is likely much worse than what one would get with three pads.

Sorry, I based my discussion on X because it is the most relevant coordinate in terms of resolution (momentum&Brho). All of them in FWHM.

> 
> One could just store the number of pads used for position reconstruction in X/Y direction with the hit. Then other detectors using the MWPC can decide which hits they want to consider. Or we could have an
> 
> ```
> enum PosMethod
> {
>     fBest,     // pick best method for every hit
>     fCoGOnly,  // always CoG
>     fSECHSOnly // set pos to NAN for less than 3 pads
> } fPosMethod;
> ```
> 
> (There are probably more modern ways to write enums.)
> 
> Also, there is some code duplication wrt x, y. I would suggest
> 
> ```
> double pos_adjust_cog(double qmax, double qneg, double qpos) 
> // in units of pitch, relative to middle strip
> {
>     return -qneg+qpos/(qmax+qneg+qpos);
> }
> 
> double pos_adjust_sechs(double qmax, double qneg, double qpos)
> {
>    // basically, return a2 for fwy=1. (should be linear in fwy)
> }
> 
> double pos_adjust(...)
> {
>    if (fPosMethod!=fCoGOnly && qpos>0 && qneg > 0)
>       return pos_adjust_sechs(...);
>    if (fPosMethod==fSECHSOnly)
>       return NAN;
>    return pos_adjust_cog(...);
> }
> 
> double GetPosY(double qmax, int padmax, double qdown, double qup)
> {
>    return padmax * fwy - (fSize / 2) + (fwy / 2) + fwy*pos_adjust(qmax, qdown, qup);
> }
> double GetPosX(double qmax, int padmax, double qleft, double qright)
> {
>    return ...;
> }
> ```
> 
> Also, I guess that S467() should be renamed so that @ManuelXarepe does not need to lie about the experiment he is analyzing, fX & friends should be local variables (std::array or std::vector), the array calData should not exist at all (an in-loop pointer to the current element replacing calData[i] is sufficient). Also ES.5. :-)

S467() and S455() are to separate between experiments where we expect one or two hits (fragments). So one could renamed them to:
- S467()   -> OneFragment() or OneHit() (or something similar)
- S455()   -> TwoFragments() or TwoHits()

(The current names were implemented in 2019-2020 thinking that these detectors would not be used in more experiments.)

ManuelXarepe:
@jose-luis-rs  Changed the name of the function from GetPositionYS522(...) to GetPositionY_COG(...).

ajedele:
Dear all, 
Could you please decide within the week what the outcome of this discussion is? Given recent proposals have been focusing more on lighter beams, I think adding the 2 pad condition is mandatory.

jose-luis-rs (approved):

jose-luis-rs:
If you are happy with the results I can merge it. Sorry for the delay.

ManuelXarepe:
I see that a PR was accepted in the meantime. I will rebase now

PR #840, opened 2023-04-05
CI/CD: Migration to Github-hosted runners for nov22

YanzhaoW:
Change to Github hosted runners for CI/CD test.

### Main features:
1. Use cvmfs inside a docker container to create an environment for the R3BRoot configuration, compilation and test.
2. Use cache and ccache to speed up the compilation process. An uncached process takes around 20 mins in total while a cached process takes only 10 mins in our case.
3. There are two caches for each PR, namely, `r3b-build-r3broot` and `r3b-build-sofia`.
4. New caches named with different SHA will be automatically created when PR is merged into the dev branch.
5. Every cache can only last 7 days unused. 
6. ctest is run with the `ctest` command.

### TODO:
1. There is a very small chance where tests for the PR are successful but tests for merging fail. We should keep a close eye on this rare situation.
2. All tests in `macros` are not run. For some reasons, they can't be run with the `ctest` command.
3. ~~All gtests also fail to run even though googletest has been installed in the environment. I will investigate this later.~~
4. Current runners only have 2 cores. With larger Github-hosted runners (4, 8, 16 cores), CI/CD process can even be faster.

---

Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 # This is a basic workflow to help you get started with Actions
-name: CI C17++
+name: CI-CD
 
 # Controls when the action will run. 
 on:
@@ -14,108 +14,63 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially
 jobs:
-  # This workflow contains a single job called "clang-format"
-  clang-format:
-    runs-on: self-hosted
+  build-test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # repos: [ r3broot, sofia, frs, asyeos, glad-tpc ]
+        repos: [ r3broot, glad-tpc, sofia-frs-asy ]
+        include:
+          - repos: r3broot
+            cache: r3broot
+          - repos: glad-tpc
+            url: https://github.com/R3BRootGroup/glad-tpc.git
+            cache: r3broot
+          - repos: sofia-frs-asy
+            url: >-
+              https://github.com/R3BRootGroup/sofia.git
+              https://github.com/R3BRootGroup/frs.git
+              https://github.com/R3BRootGroup/asyeos.git
+            cache: other-repos
 
+    container: 
+      image: yanzhaowang/cvmfs_clang:v15
+      volumes:
+        - /tmp:/cvmfs
+      env:
+        CVMDIR: /cvmfs/fairsoft.gsi.de
+        NUM_THREADS: 2
+      options: --user root --privileged  --ulimit nofile=10000:10000 --cap-add SYS_ADMIN --device /dev/fuse
     steps:
       - uses: actions/checkout@v3
+        with:
+         fetch-depth: 0
 
-      - name: Code formatting
+      - name: pre-build
+        uses: './.github/actions/pre-build'
+
+      - name: caching
+        id: caching
+        uses: './.github/actions/cache-steps'
+        with: 
+          r3b-dev-SHA: ${{ env.cacheSHA }}
+          # repo: ${{ matrix.repos }}
+          repo: ${{ matrix.cache }}
+
+      - name: build r3broot
+        uses: './.github/actions/r3bbuild-steps'
+        with:
+          ucesb-cache-existed: ${{ steps.caching.outputs.ucesb-hit }}
+
+      - name: ctest
+        # run: source $GITHUB_WORKSPACE/Dart.sh Experimental $GITHUB_WORKSPACE/.githubfiles/dart_github_CICD.cfg
+        if: ${{ github.event_name == 'pull_request' }}
         run: |
-          cd $GITHUB_WORKSPACE
-          singularity exec /data.local2/kresan/debian10-apr22-v18.6.8-ucesb.sif $PWD/util/clang-format-check.sh clang-format-8
-  # Single job called "build-r3broot"
-  build-r3broot:
-    # The type of runner that the job will run on
-    runs-on: self-hosted
-    needs: clang-format
-    if: always()
+          source $GITHUB_WORKSPACE/util/generate_geo_test.sh
+          ctest --test-dir ${BUILDDIR} --output-on-failure --rerun-failed -j${NUM_THREADS} -E "(run_gen_sim|run_digi|run_aladin_digi|califasim2|landreco|elsim)"
+        shell: bash
 
-    # Steps represent a sequence of tasks that will be executed as part of the job
-    steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v3
-
-      # Runs a set of commands using the runners shell
-      - name: Run Dart.sh
-        run: |
-          cd $GITHUB_WORKSPACE
-          git clone -b dev https://github.com/R3BRootGroup/macros.git
-          singularity exec /data.local2/kresan/debian10-apr22-v18.6.8-ucesb.sif $PWD/.githubfiles/run_build_test_apr22.sh
-  # Single job called "build-r3broot-sofia"
-  build-r3broot-sofia:
-    # The type of runner that the job will run on
-    runs-on: self-hosted
-    needs: build-r3broot
-    if: always()
-
-    # Steps represent a sequence of tasks that will be executed as part of the job
-    steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v3
-
-      # Runs a set of commands using the runners shell
-      - name: Run Dart.sh
-        run: |
-          cd $GITHUB_WORKSPACE
-          git clone -b dev https://github.com/R3BRootGroup/macros.git
-          git clone -b dev https://github.com/R3BRootGroup/sofia.git
-          singularity exec /data.local2/kresan/debian10-apr22-v18.6.8-ucesb.sif $PWD/.githubfiles/run_build_test_apr22.sh
-  # Single job called "build-r3broot-asyeos"
-#  build-r3broot-asyeos:
-    # The type of runner that the job will run on
-#    runs-on: self-hosted
-#    needs: build-r3broot
-#    if: always()
-
-    # Steps represent a sequence of tasks that will be executed as part of the job
-#    steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-#      - uses: actions/checkout@v3
-
-      # Runs a set of commands using the runners shell
-#      - name: Run Dart.sh
-#        run: |
-#          cd $GITHUB_WORKSPACE
-#          git clone -b dev https://github.com/R3BRootGroup/macros.git
-#          git clone -b dev https://github.com/R3BRootGroup/asyeos.git
-#          singularity exec /data.local2/kresan/debian10-apr22-v18.6.8-ucesb.sif $PWD/.githubfiles/run_build_test_apr22.sh
-  # Single job called "build-r3broot-glad-tpc"
-#  build-r3broot-glad-tpc:
-    # The type of runner that the job will run on
-#    runs-on: self-hosted
-#    needs: build-r3broot
-#    if: always()
-
-    # Steps represent a sequence of tasks that will be executed as part of the job
-#    steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-#      - uses: actions/checkout@v3
-
-      # Runs a set of commands using the runners shell
-      # git clone -b dev https://github.com/R3BRootGroup/glad-tpc.git
-#      - name: Run Dart.sh
-#        run: |
-#          cd $GITHUB_WORKSPACE
-#          git clone -b dev https://github.com/R3BRootGroup/macros.git
-#          singularity exec /data.local2/kresan/debian10-apr22-v18.6.8-ucesb.sif $PWD/.githubfiles/run_build_test_apr22.sh
-#  # Single job called "build-r3broot-frs"
-  build-r3broot-frs:
-    # The type of runner that the job will run on
-    runs-on: self-hosted
-    needs: build-r3broot-sofia
-    if: always()
-
-    # Steps represent a sequence of tasks that will be executed as part of the job
-    steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v3
-
-      # Runs a set of commands using the runners shell
-      - name: Run Dart.sh
-        run: |
-          cd $GITHUB_WORKSPACE
-          git clone -b dev https://github.com/R3BRootGroup/macros.git
-          git clone -b dev https://github.com/R3BRootGroup/frs.git
-          singularity exec /data.local2/kresan/debian10-apr22-v18.6.8-ucesb.sif $PWD/.githubfiles/run_build_test_apr22.sh
+      - name: ccache stats
+        if: ${{ github.event_name == 'pull_request' }}
+        run: ccache --show-stats
--- a/.github/actions/cache-steps/action.yml
+++ b/.github/actions/cache-steps/action.yml
@@ -1,0 +1,36 @@
+name: cache-steps
+description: 'steps to cache R3BRoot and ucesb'
+
+
+inputs:
+  r3b-dev-SHA:
+    description: 'SHA for r3b dev cache'
+    required: true
+  repo:
+    description: 'repository name'
+    required: true
+outputs:
+  ucesb-hit:
+    description: 'whether ucesb cache is hit'
+    value: ${{steps.cache-ucesb.outputs.cache-hit}}
+  r3b-hit:
+    description: 'whether r3b cache is hit'
+    value: ${{steps.cache-r3b.outputs.cache-hit}}
+
+runs:
+  using: composite
+  steps:
+    - name: cache ucesb
+      # if: inputs.repo == 'r3broot'
+      id: cache-ucesb
+      uses: actions/cache@v3
+      with:
+        path: ${{ env.UCESB_DIR }}
+        key: ucesb-build
+
+    - name: cache r3b
+      id: cache-r3b
+      uses: actions/cache@v3
+      with:
+        path: .ccache
+        key: r3b-build-${{ inputs.repo }}-${{ inputs.r3b-dev-SHA }}
--- a/.github/actions/pre-build/action.yml
+++ b/.github/actions/pre-build/action.yml
@@ -1,0 +1,37 @@
+name: pre build
+description: 'steps before building R3BRoot'
+
+runs:
+  using: composite
+  steps:
+    - name: set env variables
+      run: |
+        git config --global --add safe.directory $GITHUB_WORKSPACE
+        echo "UCESB_DIR=$GITHUB_WORKSPACE/ucesb/" >> $GITHUB_ENV
+        export SIMPATH=${CVMDIR}/debian10/fairsoft/nov22p1
+        echo "SIMPATH=${SIMPATH}" >> $GITHUB_ENV
+        echo "FAIRROOTPATH=${CVMDIR}/debian10/fairroot/v18.8.0_fs_nov22p1" >> $GITHUB_ENV
+        echo "${SIMPATH}/bin" >> $GITHUB_PATH
+        # variables for ccache
+        echo "CCACHE_BASEDIR=${GITHUB_WORKSPACE}" >> $GITHUB_ENV
+        echo "CCACHE_DIR=${GITHUB_WORKSPACE}/.ccache" >> $GITHUB_ENV
+        echo "CCACHE_COMPRESS=true" >> $GITHUB_ENV
+        echo "CCACHE_COMPRESSLEVEL=6" >> $GITHUB_ENV
+        echo "SOURCEDIR=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+        echo "BUILDDIR=$GITHUB_WORKSPACE/build" >> $GITHUB_ENV
+        echo "cacheSHA=$([[ ${{github.event_name}} = "push" ]] && git rev-parse --short HEAD || git rev-parse --short HEAD^)" >> $GITHUB_ENV
+      shell: bash
+
+
+    - name: mount cvmfs
+      run: |
+        apt-get install -y ccache
+        wget https://cernbox.cern.ch/remote.php/dav/public-files/RmnTeeOZpYjCJQb/masterkey.gsi.de.pub
+        mv masterkey.gsi.de.pub /etc/cvmfs/keys
+        cp .githubfiles/fairsoft.gsi.de.conf /etc/cvmfs/config.d
+        mkdir -p ${CVMDIR}
+        mount -t cvmfs fairsoft.gsi.de ${CVMDIR}
+        mkdir -p ${UCESB_DIR}
+        mkdir -p ${CCACHE_DIR}
+        echo $PATH
+      shell: bash
--- a/.github/actions/r3bbuild-steps/action.yml
+++ b/.github/actions/r3bbuild-steps/action.yml
@@ -1,0 +1,49 @@
+name: build-steps
+description: 'steps to build R3BRoot and ucesb'
+
+
+inputs:
+  ucesb-cache-existed:
+    description: 'has ucesb build cache'
+    required: false
+    default: 'false'
+
+runs:
+  using: composite
+  steps:
+    - name: check
+      run: |
+        echo "${{ inputs.ucesb-cache-existed }}"
+      shell: bash
+
+    - name: build ucesb
+      if: inputs.ucesb-cache-existed != 'true'
+      run: |
+        echo "ucesb directory is set as ${UCESB_DIR}"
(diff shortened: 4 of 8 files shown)


YanzhaoW:
@jose-luis-rs @kresan @inkdot7 Any concerns about this PR?

jose-luis-rs:
Hi, for me it is fine.
Just a question related to:
cmake --fresh . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=on -DBUILD_GEOMETRY=OFF -DUSE_DIFFERENT_COMPILER=TRUE \

What is the reason to disable the generation of root geometries? did you find problems?

YanzhaoW:
> Hi, for me it is fine. Just a question related to: cmake --fresh . -B build -DCMAKE_EXPORT_COMPILE_COMMANDS=on -DBUILD_GEOMETRY=OFF -DUSE_DIFFERENT_COMPILER=TRUE \
> 
> What is the reason to disable the generation of root geometries? did you find problems?

Thanks for the feedback. Yes, it would make the compilation very slow. To speed up the whole CI/CD, only necessary geo files are created later for the test.

jose-luis-rs:
Ok, thanks for the comments.
I would like to do a small request, do you have time to include in these checks the asyeos repository?
https://github.com/R3BRootGroup/asyeos
to ensure that the modifications in R3BRoot are compatible with the Asyeos code. We will need it for future experiments...
Thanks in advance

.github/workflows/main.yml
```diff
@@ -14,108 +14,62 @@ on:
...
+            cache: r3broot
+          - repos: sofia
+            url: https://github.com/R3BRootGroup/sofia.git
+            cache: sofia
+          - repos: frs
+            url: https://github.com/R3BRootGroup/frs.git
+            cache: r3broot
+          - repos: asyeos
+            url: https://github.com/R3BRootGroup/asyeos.git
+            cache: r3broot
```
YanzhaoW: @jose-luis-rs In the future, if you wish to add more repos, you can specify them here with the corresponding url. If they are very different from the base R3BRoot, like sofia here, it could take a long time for them to build. In such case, they need their own cache to speed up the building. If not, like `frs` here, we can just use r3broot base cache for the building.

Also be very careful about the additional ctest in those repos. If those tests need some geo files, we should also create those geo files with the file `R3BRoot/uil/generate_geo_test.sh`

YanzhaoW:
Ok, Something wrong with my GTest PR #843. It's only successful if gtest is not installed in the local system but failed if it's already installed. I fixed it in this PR. Sorry for that.

.github/workflows/main.yml
jose-luis-rs: Thanks for the comments!

YanzhaoW:
@jose-luis-rs Asyeos has been added successfully in the CI. Is there anything else I need to add?

jose-luis-rs (approved):

inkdot7:
Many thanks for the work on CI !!!

hapol:
Thank you very much!! Is glad-tpc folder included?

jose-luis-rs:
> Thank you very much!! Is glad-tpc folder included?

Please, don't forget this issue:
https://github.com/R3BRootGroup/glad-tpc/issues/28

YanzhaoW:
@jose-luis-rs Ok, I made a [PR](https://github.com/R3BRootGroup/glad-tpc/pull/29) in glad-tpc to fix the compilation error and warnings. Please have a quick merge over there and restart the CI test here.

YanzhaoW:
glad-tpc is also included. 

But I suggest we shouldn't let this trend going further. R3BRoot is an upstream repo to those repos like sofia or frs. And those downstream repos should have their own CI test checked for their new PRs rather than relying on the checking from the upstream. 

The CI/CD test of FairRoot has never considered and doesn't need to consider whether the PR over there breaks R3BRoot or not. With the same logic, in my personal opinion, CI/CD test of R3BRoot doesn't need to consider whether the PR breaks the downstream repos.

But I'm all ears for different opinions.

@jose-luis-rs @hapol 

Edit: Or we could put all those tests for other repos into one single CI run?

hapol:
Thank you.
I guess that glad-tpc was out of the main R3BRoot tree due to development reasons, in part the complexity of reconstruction of gas detector, and somehow make the initial development independent if it needs some particular dependences, as it was partially derived from other projects. But now, I think that there is no reason to make it independent and sooner or later has to be integrated.

jose-luis-rs:
You are right @YanzhaoW, but in this case sofia, asyeos, frs, ... are part of R3BRoot and used for the analysis of some R3B experiments, so somehow we have to ensure that the combination of all these repositories works. 

In the case of glad-tpc, it is a single detector and its integration in R3BRoot is easy. In the case of the other repositories this integration is more complicated because they contain several detectors and some of them could not be used anymore.

YanzhaoW:
> are part of R3BRoot and used for the analysis of some R3B experiments, so somehow we have to ensure that the combination of all these repositories works.

No problem. Then if it's ok, I will put all those experiment repos in a single run.

YanzhaoW:
Now sofia, frs and asyeos are downloaded together and checked together. glad-tpc is still checked alone since it will be integrated into r3broot dev sooner or later.

inkdot7:
@YanzhaoW I'm not sure if @hapol meant that they (sofia/frs/asy) need to be checked together with each other.  That would be if they are used together, and could interfere with each other in that case.  Just that they need to be checked together with r3broot.  If the separate testing was faster using more jobs, that would be an advantage.  It can also be nice to know which parts are broken when not all of them.

Either way, a way to choose a specific commit for those dependent repositories is needed:

- The combination of r3broot with such experiment/setup-specific codes needs to be ensured.  That is one of the main purposes of the CI testing.  I.e. to ensure that updates to r3broot are accompanied by the necessary updates in the experiment/setup repositories.
- At the same time, r3broot development must be possible, while maintaining an 'always-green-CI' policy.  Thus changes which require updates in the dependent repositories need to be possible, by marking which (new) commit of a dependent repository is used to perform the successful build.

One way would be to include them (the dependents) as git submodules at least for this CI purpose.  Does not mean that they have to be used that way otherwise.  But that would give a standard way of declaring specific commits.

YanzhaoW:
> I'm not sure if @hapol meant that they (sofia/frs/asy) need to be checked together with each other. That would be if they are used together, and could interfere with each other in that case. Just that they need to be checked together with r3broot.

I kinda prefer they are checked together for some reasons:

1. The implementation of each experiment should not influence the implementation of other experiments. And R3BRoot dev should also not be configured to suit for specific experiment. Which means, in principle, running all experiments at the same time should not be a problem. If there is a problem to do this, something must be not quite right and we should fix that. And testing separately with R3BRoot could also be done in the downstream repos. 

2. Each run has a cache to speed up the building process. The total size of cache files is limited to 10 GB by Github. Running each experiment in separate runs can quickly deplete our cache space if there are too many PR merged in a very short time.

 > If the separate testing was faster using more jobs, that would be an advantage. It can also be nice to know which parts are broken when not all of them.

I also had similar thought. But later I think this shouldn't be a problem because if there is an error we need to track down its source file anyway.

>Thus changes which require updates in the dependent repositories need to be possible, by marking which (new) commit of a dependent repository is used to perform the successful build.

I don't quite understand your point. The current CI in this PR just download the latest commit in dev branch. What could be possible problem for this?

inkdot7:
> And testing separately with R3BRoot could also be done in the downstream repos.

Please consider R3BRoot and (sofia/frs/asy/...) as being logically part of the _same_ repository.  What the CI should achieve is to make it unlikely that *any* change (merge request) breaks the experiment analysis (even for short times!) for users that regularly update their repositories.  I.e. the merger (here @jose-luis-rs) must know that there is an already prepared update for all dependent repositories which are affected.  They way to know that is that the CI signals 'passed'.

If there is not a way to specify different commits for dependent repositories, then it becomes impossible* to make R3BRoot developments that involve e.g. prototype changes in functions, or name changes or significant structure changes.  (Or any other more subtle change in the code that does not break compilation, but affects result.  E.g. some different sorting of entries in arrays can in some cases affect the overall behaviour of routines.)

\* Impossible while keeping a CI-shall-be-green-for-merge policy.

> The current CI in this PR just download the latest commit in dev branch. What could be possible problem for this?

Chicken-egg-problem.  The dependent repository cannot merge a fix until R3BRoot has merged to the new behaviour.

One way which makes this possible is git submodules.  As the reference changes together with the 'breaking' update.

The CI would eventually not only test compilation of various setups, but also run many more small tests to check that the data treatment stays compatible.  I suppose this is already done to some extent, but imagine it covering many more aspects.  The more diverse things that are tested, the easier bugs are found.  The most diverse test cases for R3BRoot is the actual different setups/experiments.

YanzhaoW:
@inkdot7 Thanks very much for your comment.

> Please consider R3BRoot and (sofia/frs/asy/...) as being logically part of the same repository. What the CI should achieve is to make it unlikely that any change (merge request) breaks the experiment analysis (even for short times!) for users that regularly update their repositories. I.e. the merger (here @jose-luis-rs) must know that there is an already prepared update for all dependent repositories which are affected. They way to know that is that the CI signals 'passed'.

Yes, I also think so.

>If there is not a way to specify different commits for dependent repositories, then it becomes impossible* to make R3BRoot developments that involve e.g. prototype changes in functions, or name changes or significant structure changes. (Or any other more subtle change in the code that does not break compilation, but affects result. E.g. some different sorting of entries in arrays can in some cases affect the overall behaviour of routines.)

I not sure why it becomes impossible if there is no way so specify different commits for dependent repositories. Or maybe I don't quite understand your point very well. In my opinion, the developments involving "any other more subtle change in the code that does not break compilation, but affects result." should be checked with **unit test**. And I can't see any relevance of R3BRoot CI being able to specify different commits of dependent repos.

>Chicken-egg-problem. The dependent repository cannot merge a fix until R3BRoot has merged to the new behaviour.

I don't quite understand this neither. The CI process of R3BRoot can only be triggered with PRs in R3BRoot repo and the CI processes of dependent repos can only be triggered with their own PRs (unless there is a cross-repo triggering which I'm not aware of). Therefore, whether a PR in dependent repo can be merged or not can only be decided by its own CI instead of the CI in R3BRoot repo. If my understanding is correct, I feel it's kind of bad to have a PR in a dependent repo which requires another PR (not merged) in R3BRoot repo. Instead, a PR in a dependent repo should only be made with respect to the latest already existing version of R3BRoot. 

>One way which makes this possible is git submodules. As the reference changes together with the 'breaking' update.

I'm not very familiar with git submodules. Though I do think this's a better way to download dependencies. The current method in this CI is to download the the latest head of dev branch of each dependency for each run:
```sh
git clone -b dev ${url}
```  
With this, the reference should also change together with the 'breaking' update from dependencies (if there is a update in dev branch). Sorry I still can't see what the problem could be.

>The CI would eventually not only test compilation of various setups, but also run many more small tests to check that the data treatment stays compatible. I suppose this is already done to some extent, but imagine it covering many more aspects. The more diverse things that are tested, the easier bugs are found. The most diverse test cases for R3BRoot is the actual different setups/experiments.

I think you are talking about unit test? (Please correct me if I'm wrong.) If so, I agree that we need more unit tests for each detector and setup. But I seriously doubt that most people can manage to do this since many of us aren't even willing to make the code in the PR clean, let alone have unit tests.

inkdot7:
@YanzhaoW Thanks for the update.  I think this is very important.  So am happy to discuss and try to explain better what I'm after.  Towards the end is something I marked as 'GOAL'.  This is what in my mind drives the rest.  Also important is I think the relationship distinction two paragraphs down.

My point essentially is:  R3BRoot cannot take the approach that it is an upstream repository, to which the experiment repositories have to adapt themselves.

The *relationship* between R3BRoot and experiment repositories is not the same as between FairRoot and R3BRoot:

Changes to R3BRoot that require changes in the experiment directories, must be accompanied by those changes.  CI is the guardian of that.

There seem to be three ways of doing that:

1. Put the experiment directories directly in the R3BRoot repository.
  Then any PR request that would break an experiment directory would have to also has fix that directory, since CI fails. Being in the same repository makes this straightforward.
  Except for using this 'thought approach' as an explanation I am not suggesting to merge the repositories.  But CI should behave 'as if'.

2. Using another means of ensuring the same thing.
  I'm not overly liking git submodules.  But they are one way which can achieve the effect. 

3. Breaking changes have to update the CI download lines `git clone -b BRANCH_NAME_WITH_FIXES ${url}`.
  This then stays in R3BRoot until 'dev' branch of those repositories have merged the needed changes.

I am aware that not many write units tests.   But if someone does that work in order to ensure that R3BRoot does not break his/her experiment, then that effort should pay off.  __*GOAL*__: I.e. PR requests that break those CI (unit) tests do not get merged until fixed.  And it should be possible to place such tests inside an experiment directory/repository, as that has the infrastructure to process (test) data of that experiment.

(That of course does not mean to just let those tests rot away.  Testing infrastructure requires maintenance.)

jose-luis-rs:
> @YanzhaoW Thanks for the update. I think this is very important. So am happy to discuss and try to explain better what I'm after. Towards the end is something I marked as 'GOAL'. This is what in my mind drives the rest. Also important is I think the relationship distinction two paragraphs down.
> 
> My point essentially is: R3BRoot cannot take the approach that it is an upstream repository, to which the experiment repositories have to adapt themselves.
> 
> The _relationship_ between R3BRoot and experiment repositories is not the same as between FairRoot and R3BRoot:
> 
> Changes to R3BRoot that require changes in the experiment directories, must be accompanied by those changes. CI is the guardian of that.
> 
> There seem to be three ways of doing that:
> 
>     1. Put the experiment directories directly in the R3BRoot repository.
>        Then any PR request that would break an experiment directory would have to also has fix that directory, since CI fails. Being in the same repository makes this straightforward.
>        Except for using this 'thought approach' as an explanation I am not suggesting to merge the repositories.  But CI should behave 'as if'.
> 
>     2. Using another means of ensuring the same thing.
>        I'm not overly liking git submodules.  But they are one way which can achieve the effect.
> 
>     3. Breaking changes have to update the CI download lines `git clone -b BRANCH_NAME_WITH_FIXES ${url}`.
>        This then stays in R3BRoot until 'dev' branch of those repositories have merged the needed changes.
> 
> 
> I am aware that not many write units tests. But if someone does that work in order to ensure that R3BRoot does not break his/her experiment, then that effort should pay off. **_GOAL_**: I.e. PR requests that break those CI (unit) tests do not get merged until fixed. And it should be possible to place such tests inside an experiment directory/repository, as that has the infrastructure to process (test) data of that experiment.
> 
> (That of course does not mean to just let those tests rot away. Testing infrastructure requires maintenance.)

Thanks for the discussion.

Initially the CIs for the frs, sofia, asyeos, ... can be checked together because there are not interferences among them. If one of them fails during the current CI implemented by @YanzhaoW means that it also fails in a separate test.

Personally, I don't like the use of submodules with concurrently changing repositories. I used it in the past for some Geant4 development and was a disaster.

Dependencies:
- The frs repository was only used for some tests in 2019 and the detectors implemented there (old FRS-Musics, MWs, TPCs,...) will not used for FAIR. So in a couple of years (2-3) this repository will be completely obsolete. But at the moment it is better to ensure that it can compile with R3BRoot.

- The sofia repository is used for the analysis of some experiments performed in 2020 and 2021 and some thesis and publications depend on it. So this repository is very important. The link with R3BRoot is through MWPCs and MUSIC detectors (twim&music). So we have to be sure that the modifications in these detectors don't change the results. 

- The asyeos repository is linked to R3BRoot through the LOS and NeuLAND detectors, but at the moment the analysis part is not implemented yet. In the future, we will need to create some unit tests to check that the start time and the correlations with NeuLAND are right.

We have CI tests in all the repositories to check that our PRs do not break the compilation. Unit tests should be only in the main repository (R3bRoot) because it contains all the needed detectors. At the moment we only have unit tests for simulations, but this kind of test is also needed for the analysis part. So we should create a github repository to put there some lmd files with a few events to run some examples of analysis macros and check that the information (detector output branches, correlations, ...) is properly stored in the output root files. I already have some of these tests running in our cluster for the fission experiments, but I didn't have time to put them in github. Sorry for that.

@YanzhaoW, if you want, I could create a github repository for the lmd files and parameter files needed for the data analysis unit tests and then you could create in R3BRoot the corresponding unit test to check some reference outputs for NeuLAND (cosmic rays, time correlations with LOS, ...). Of course, if you have time!

YanzhaoW:
@inkdot7  Thanks for your tips about git submodules.

I looked up a bit about git submodules. And to be honest, I find it's probably not very suitable for our case.

First we need to recognise there does exist a dependency between R3BRoot and, for example, sofia. This is due to a simple reason that R3BRoot can be used alone without sofia or other experiment specific repos while sofia  must be used with R3BRoot. In other words, sofia repo is relying/dependent on R3BRoot repo, no the other way around.

Now about git submodule. Here is a statement that I found in internet:

> In most cases, Git submodules are used when your project becomes more complex, and while your project depends on the main Git repository, you might want to keep their change history separate.

So if we have two projects, let's say, A and B and if A is dependent on B and A cannot be compiled/used without B, we should make project B as a git submodule for project A. In our case, the correct way to use git submodule is to make R3BRoot project a submodule for the project sofia, not the other way around.

Another way to view this matter is to see whether A need to be used with an older version of B or whether B need to be used with an older version of A. So here I could think a case where latest version of sofia may need to work with older version of R3BRoot (for example, if R3BRoot moves to C++20 and people working with sofia don't like C++20. In this case, they need to use the older version R3BRoot.). But I can't imagine a case where the latest version of R3BRoot need to be used with an older version of sofia. Please correct me if I'm wrong. (@jose-luis-rs )

> Changes to R3BRoot that require changes in the experiment directories, must be accompanied by those changes.

If I understand you correctly, I don't think this case should be allowed. Changes to R3BRoot should not require changes in a specific experiment directory. Otherwise, we would make R3BRoot dependent on those experiments and it could be very nasty when we adapt the R3BRoot to the future experiments. (again, please correct me if I'm wrong.)

> Then any PR request that would break an experiment directory would have to also has fix that directory, since CI fails. Being in the same repository makes this straightforward.
Except for using this 'thought approach' as an explanation I am not suggesting to merge the repositories. But CI should behave 'as if'.

I'm not a fan of putting all things together in one place. We should instead follow unix philosophy "Do one thing and do it well". But with git submodule, I don't think it could even work. Suppose we use git submodules for sofia, frs or asyeos, when I fork the R3BRoot to my own github account, I can't change anything inside sofia or frs. This is because in .gitmodules, those sub folders are linked directly to the repos from R3BRootGroup, not from my forked sofia or frs. Therefore, I can't do any modifications for those submodules in my R3BRoot pull request.

inkdot7:
>> Changes to R3BRoot that require changes in the experiment directories, must be accompanied by those changes.
>
> If I understand you correctly, I don't think this case should be allowed. Changes to R3BRoot should not require changes in a specific experiment directory. Otherwise, we would make R3BRoot dependent on those experiments and it could be very nasty when we adapt the R3BRoot to the future experiments. (again, please correct me if I'm wrong.)

This would mean that if there is a function `foo (int *something);` inside R3BRoot and that is used in code in e.g. sofia repository, and one then would like to change it to `foo (double *something);`, then one can never do that, since it is not allowed to make a change in R3BRoot which breaks sofia.

jose-luis-rs:
> @inkdot7 Thanks for your tips about git submodules.
> 
> I looked up a bit about git submodules. And to be honest, I find it's probably not very suitable for our case.
> 
> First we need to recognise there does exist a dependency between R3BRoot and, for example, sofia. This is due to a simple reason that R3BRoot can be used alone without sofia or other experiment specific repos while sofia must be used with R3BRoot. In other words, sofia repo is relying/dependent on R3BRoot repo, no the other way around.
> 
> Now about git submodule. Here is a statement that I found in internet:
> 
> > In most cases, Git submodules are used when your project becomes more complex, and while your project depends on the main Git repository, you might want to keep their change history separate.
> 
> So if we have two projects, let's say, A and B and if A is dependent on B and A cannot be compiled/used without B, we should make project B as a git submodule for project A. In our case, the correct way to use git submodule is to make R3BRoot project a submodule for the project sofia, not the other way around.
> 
> Another way to view this matter is to see whether A need to be used with an older version of B or whether B need to be used with an older version of A. So here I could think a case where latest version of sofia may need to work with older version of R3BRoot (for example, if R3BRoot moves to C++20 and people working with sofia don't like C++20. In this case, they need to use the older version R3BRoot.). But I can't imagine a case where the latest version of R3BRoot need to be used with an older version of sofia. Please correct me if I'm wrong. (@jose-luis-rs )

R3BRoot-asyeos and R3BRoot-sofia work as full analysis packages, if R3BRoot moves to C++20 the other repositories must also do the same update because they could require the latest version of R3BRoot for analysis. Moreover we could have serious problems to run future experiments if these repositories cannot run together. Another important point is the manpower to do all these modifications.

> 
> > Changes to R3BRoot that require changes in the experiment directories, must be accompanied by those changes.
> 
> If I understand you correctly, I don't think this case should be allowed. Changes to R3BRoot should not require changes in a specific experiment directory. Otherwise, we would make R3BRoot dependent on those experiments and it could be very nasty when we adapt the R3BRoot to the future experiments. (again, please correct me if I'm wrong.)
> 
> > Then any PR request that would break an experiment directory would have to also has fix that directory, since CI fails. Being in the same repository makes this straightforward.
> > Except for using this 'thought approach' as an explanation I am not suggesting to merge the repositories. But CI should behave 'as if'.
> 
> I'm not a fan of putting all things together in one place. We should instead follow unix philosophy "Do one thing and do it well". But with git submodule, I don't think it could even work. Suppose we use git submodules for sofia, frs or asyeos, when I fork the R3BRoot to my own github account, I can't change anything inside sofia or frs. This is because in .gitmodules, those sub folders are linked directly to the repos from R3BRootGroup, not from my forked sofia or frs. Therefore, I can't do any modifications for those submodules in my R3BRoot pull request.

YanzhaoW:
> This would mean that if there is a function foo (int *something); inside R3BRoot and that is used in code in e.g. sofia repository, and one then would like to change it to foo (double *something);, then one can never do that, since it is not allowed to make a change in R3BRoot which breaks sofia.

I think it's not impossible to have R3BRoot compatible to both older and latest version of sofia. In this example, we could declare both two functions with same name (function overloading in C++) first in the R3BRoot. After it being merged to dev branch, we then use the new function in sofia. After some time, when nobody really needs that function (`foo(int* something)`) anymore, we can deprecate it in R3BRoot.

YanzhaoW:
> if you want, I could create a github repository for the lmd files and parameter files needed for the data analysis unit tests and then you could create in R3BRoot the corresponding unit test to check some reference outputs for NeuLAND (cosmic rays, time correlations with LOS, ...). Of course, if you have time!

Thanks. I certainly would like to have some lmd files to test on (when I begin to rewrite Neuland calibration process). But I would say a github repo would not be a good choice to store data. Rather, Zenodo could be a much better option.

inkdot7:
> I think it's not impossible to have R3BRoot compatible to both older and latest version of sofia. In this example, we could declare both two functions with same name (function overloading in C++) first in the R3BRoot. After it being merged to dev branch, we then use the new function in sofia. After some time, when nobody really needs that function (foo(int* something)) anymore, we can deprecate it in R3BRoot.

You are sure that all reasonable changes can be handled like that?  E.g. change a structure member name which is used in an experiment directory, and then overloading does not work.

I would suggest to merge these CI tests, and then see how it works in practice.

YanzhaoW:
@inkdot7 Ok, yeah, I understand your point. But in your case, it's still possible to maintain a short-term backward compatibility: Rather than change the name of the member variable, we could keep the old variable and add another variable (or reference to the older variable) with a new name. I agree it's nasty.

So let's merge this PR and see how it works in practice. @jose-luis-rs

jose-luis-rs:
> @inkdot7 Ok, yeah, I understand your point. But in your case, it's still possible to maintain a short-term backward compatibility: Rather than change the name of the member variable, we could keep the old variable and add another variable (or reference to the older variable) with a new name. I agree it's nasty.
> 
> So let's merge this PR and see how it works in practice. @jose-luis-rs

Yes, just a last comment because now we are not sending the CI results to the CDash framework:
https://cdash.gsi.de/index.php?project=R3BRoot
Is this still possible?

This helps to see the evolution with time and the problems in general:
https://cdash.gsi.de/index.php?project=FairRoot

YanzhaoW:
@jose-luis-rs No problem at all. Maybe I could do this in another PR? But is this urgent?

jose-luis-rs:
> @jose-luis-rs No problem at all. Maybe I could do this in another PR? But is this urgent?

Ok, it is not urgent.

inkdot7:
@jose-luis-rs @YanzhaoW  I'm not sure where to report this, but after rebasing #712 it fails CI (static analysis / clang-tidy) with an error which I cannot see has anything to do with that commit:
```
/__w/R3BRoot/R3BRoot/r3bsource/base/R3BReader.h:22:10: error: 'ext_data_struct_info.hh' file not found [clang-diagnostic-error]
#include "ext_data_struct_info.hh"
         ^~~~~~~~~~~~~~~~~~~~~~~~~
```
For some reason, I do not see the CI results of this PR (i.e. #840) any longer, but assume that they all passed when it was merged.  Given that the current commit for #712 is just a commit following this merged #840, it is quite confusing.

(The other three CI tests did pass for #712)

PR #842, opened 2023-04-11
Fibers and ToFD calibration, parameter and Tracking update.

enlorenz:
Fibers and ToFD calibration, parameter and Tracking update.

Fibers:
Adjusted parameter calculation and application. Minor changes, ToF and ToT calculation now consistent.

ToFD:
Cal2Hit adjusted to be used with new parameter structure. Adjusted mathematical operations, and introduced some failsafes.
Code is not "clean" yet, but needed so that the parameters can be used for further analysis.
Cal2HitPar is in a "running" state and delivers parameters for "s522/s509". Not advised to use for other experiments yet. Code needs a proper re-work, before it is safe for general use.
Individual calibration steps/functions with minor changes. introduced higher order correction step using physics data.

Fragment tracker:
Standalone task. Uses Fibers and ToFD to track fragments behind GLAD and calculate in real time observables and deliver PID. Still under development.
---

Checklist:

* [ ] Rebased against `dev` branch
* [ ] My name is in the resp. CONTRIBUTORS/AUTHORS file
* [ ] Followed [the seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/tofd/calibration/R3BTofDCal2Hit.cxx
+++ b/tofd/calibration/R3BTofDCal2Hit.cxx
@@ -1,6 +1,6 @@
 /******************************************************************************
  *   Copyright (C) 2019 GSI Helmholtzzentrum für Schwerionenforschung GmbH    *
- *   Copyright (C) 2019-2023 Members of R3B Collaboration                     *
+ *   Copyright (C) 2019 Members of R3B Collaboration                          *
  *                                                                            *
  *             This software is distributed under the terms of the            *
  *                 GNU General Public Licence (GPL) version 3,                *
@@ -26,13 +26,14 @@
 #include "R3BTofDMappingPar.h"
 #include "R3BTofdCalData.h"
 #include "R3BTofdHitData.h"
-#include <FairRootManager.h>
 
+#include <FairRootManager.h>
 #include "FairLogger.h"
 #include "FairRuntimeDb.h"
 #include "TH1F.h"
 #include "TH2F.h"
 #include "THnSparse.h"
+#include "TFile.h"
 
 #include "TClonesArray.h"
 #include "TMath.h"
@@ -155,10 +156,10 @@ R3BTofDCal2Hit::~R3BTofDCal2Hit()
 
 void R3BTofDCal2Hit::SetParContainers()
 {
-    fMapPar = dynamic_cast<R3BTofDMappingPar*>(FairRuntimeDb::instance()->getContainer("tofdMappingPar"));
-    R3BLOG_IF(warn, !fMapPar, "Could not get access to tofdMappingPar container");
+    fMapPar = (R3BTofDMappingPar*)FairRuntimeDb::instance()->getContainer("tofdMappingPar");
+    R3BLOG_IF(warning, !fMapPar, "Could not get access to tofdMappingPar container");
 
-    fHitPar = dynamic_cast<R3BTofDHitPar*>(FairRuntimeDb::instance()->getContainer("tofdHitPar"));
+    fHitPar = (R3BTofDHitPar*)FairRuntimeDb::instance()->getContainer("tofdHitPar");
     if (!fHitPar)
     {
         R3BLOG(error, "Could not get access to tofdHitPar container");
@@ -191,13 +192,13 @@ InitStatus R3BTofDCal2Hit::Init()
         return kFATAL;
     }
 
-    header = dynamic_cast<R3BEventHeader*>(mgr->GetObject("EventHeader."));
+    header = (R3BEventHeader*)mgr->GetObject("EventHeader.");
     R3BLOG_IF(fatal, NULL == header, "EventHeader. not found");
 
-    fCalItems = dynamic_cast<TClonesArray*>(mgr->GetObject("TofdCal"));
+    fCalItems = (TClonesArray*)mgr->GetObject("TofdCal");
     R3BLOG_IF(fatal, NULL == fCalItems, "TofdCal not found");
 
-    fCalTriggerItems = dynamic_cast<TClonesArray*>(mgr->GetObject("TofdTriggerCal"));
+    fCalTriggerItems = (TClonesArray*)mgr->GetObject("TofdTriggerCal");
     R3BLOG_IF(fatal, NULL == fCalTriggerItems, "TofdTriggerCal not found");
 
     maxevent = mgr->CheckMaxEventNo();
@@ -317,7 +318,7 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
     // puts("Event");
     for (Int_t ihit = 0; ihit < nHits; ihit++)
     {
-        auto* hit = dynamic_cast<R3BTofdCalData*>(fCalItems->At(ihit));
+        auto* hit = (R3BTofdCalData*)fCalItems->At(ihit);
         size_t idx = (hit->GetDetectorId() - 1) * fPaddlesPerPlane + (hit->GetBarId() - 1);
 
         // std::cout << "Hits: " << hit->GetDetectorId() << ' ' << hit->GetBarId() << ' ' << hit->GetSideId() << '  '
@@ -334,7 +335,7 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
     std::vector<R3BTofdCalData const*> trig_map;
     for (int i = 0; i < fCalTriggerItems->GetEntriesFast(); ++i)
     {
-        auto trig = dynamic_cast<R3BTofdCalData const*>(fCalTriggerItems->At(i));
+        auto trig = (R3BTofdCalData const*)fCalTriggerItems->At(i);
         if (trig_map.size() < trig->GetBarId())
         {
             trig_map.resize(trig->GetBarId());
@@ -388,9 +389,9 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
             {
                 if (!s_was_trig_missing)
                 {
-                    R3BLOG(error, "Missing trigger information!");
-                    R3BLOG(error, "Top: " << top->GetDetectorId() << ' ' << top->GetSideId() << ' ' << top->GetBarId());
-                    R3BLOG(error, "Bot: " << bot->GetDetectorId() << ' ' << bot->GetSideId() << ' ' << bot->GetBarId());
+                    //R3BLOG(error, "Missing trigger information!");
+                    //R3BLOG(error, "Top: " << top->GetDetectorId() << ' ' << top->GetSideId() << ' ' << top->GetBarId());
+                    //R3BLOG(error, "Bot: " << bot->GetDetectorId() << ' ' << bot->GetSideId() << ' ' << bot->GetBarId());
                     s_was_trig_missing = true;
                 }
                 ++n2;
@@ -434,10 +435,24 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
                 auto top_tot = fmod(top->GetTimeTrailing_ns() - top->GetTimeLeading_ns() + c_range_ns, c_range_ns);
                 auto bot_tot = fmod(bot->GetTimeTrailing_ns() - bot->GetTimeLeading_ns() + c_range_ns, c_range_ns);
 
-                auto THit_raw = (bot->GetTimeLeading_ns() + top->GetTimeLeading_ns()) / 2.; // needed for TOF for ROLUs
+               // auto THit_raw = (bot->GetTimeLeading_ns() + top->GetTimeLeading_ns()) / 2.; // needed for TOF for ROLUs
+                auto THit_raw = fTimeStitch->GetTime((bot_trig_ns + top_trig_ns) / 2. - header->GetTStart(),"tamex","vftx"); // needed for TOF for ROLUs
 
-                // std::cout<<"ToT: "<<top_tot << " "<<bot_tot<<"\n";
-
+                //std::cout<<"ToT: "<<top_tot << " "<<bot_tot<<"\n";
+                 //std::cout<<"BarT: "<<top->GetBarId()<<" BarB: "<<bot->GetBarId()<<endl;
+                if(top->GetBarId() == bot->GetBarId()){ 
+                    bool both = false;
+                    nev++;
+                    if(top_tot > c_range_ns/10 && bot_tot > c_range_ns/10){ nev_lmboth++; both = true;}
+                    if(top_tot > c_range_ns/10){ 
+                        top_tot = bot_tot;
+                        if(!both) nev_lmt++;
+                    }
+                    if(bot_tot > c_range_ns/10){
+                         bot_tot = top_tot;
+                         if(!both) nev_lmb++;
+                    }
+                }
                 // register multi hits
                 vmultihits[iPlane][iBar] += 1;
 
@@ -448,6 +463,8 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
                     continue;
                 }
 
+                Double_t Ctop_tot = top_tot * par->GetToTOffset2();
+                Double_t Cbot_tot = bot_tot * par->GetToTOffset1();
                 // walk corrections
                 if (par->GetPar1Walk() == 0. || par->GetPar2Walk() == 0. || par->GetPar3Walk() == 0. ||
                     par->GetPar4Walk() == 0. || par->GetPar5Walk() == 0.)
@@ -456,13 +473,13 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
                 }
                 else
                 {
-                    auto bot_ns_walk = bot_ns - walk(bot_tot,
+                    auto bot_ns_walk = bot_ns - walk(Cbot_tot,
                                                      par->GetPar1Walk(),
                                                      par->GetPar2Walk(),
                                                      par->GetPar3Walk(),
                                                      par->GetPar4Walk(),
                                                      par->GetPar5Walk());
-                    auto top_ns_walk = top_ns - walk(top_tot,
+                    auto top_ns_walk = top_ns - walk(Ctop_tot,
                                                      par->GetPar1Walk(),
                                                      par->GetPar2Walk(),
                                                      par->GetPar3Walk(),
@@ -475,6 +492,7 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
 
                 // calculate time of hit
                 Double_t THit = (bot_ns + top_ns) / 2. - par->GetSync();
+          ///      cout<<"GetSynch "<<par->GetSync()<<" tb "<<bot_ns<<" tt "<<top_ns<<" THit_raw "<<(bot_ns + top_ns) / 2.<<" THit "<<(bot_ns + top_ns) / 2. - par->GetSync()<<endl;
                 if (std::isnan(THit))
                 {
                     R3BLOG(fatal, "TofD THit not found");
@@ -487,7 +505,9 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
 
                 // calculate y-position from ToT
                 auto posToT =
-                    par->GetLambda() * log((top_tot * par->GetToTOffset2()) / (bot_tot * par->GetToTOffset1()));
+                    par->GetLambda() * log((Ctop_tot) / (Cbot_tot));
+                //    par->GetLambda() * log((top_tot) / (bot_tot));
+        ///            cout<<"Lambda "<<par->GetLambda()<<" GetVeff "<<par->GetVeff()<<endl;
 
                 if (fTofdTotPos)
                 {
@@ -516,7 +536,7 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
 
                 Double_t para[4];
                 Double_t qb = 0.;
-                if (fTofdQ > 0)
+                if (fTofdQ > 0 /*&& 1*/)
                 {
                     if (fTofdTotPos)
                     {
@@ -525,8 +545,9 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
                         para[1] = par->GetPolb();
                         para[2] = par->GetPolc();
                         para[3] = par->GetPold();
-                        qb = TMath::Sqrt(top_tot * bot_tot) /
-                             (para[0] + para[1] * pos + para[2] * pow(pos, 2) + para[3] * pow(pos, 3));
+                 //       qb = TMath::Sqrt(Ctop_tot * Cbot_tot) /
+                        qb = TMath::Sqrt(Ctop_tot * Cbot_tot) *
+                             (para[0] + para[1] * pos + para[2] * pow(pos, 2) + para[3] * pow(pos, 3)/*1.*/);
                         qb = qb * fTofdQ;
                     }
                     else
@@ -536,13 +557,13 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
                         para[1] = par->GetPar1b();
                         para[2] = par->GetPar1c();
                         para[3] = par->GetPar1d();
-                        auto q1 = bot_tot /
+                        auto q1 = Cbot_tot /
                                   (para[0] * (exp(-para[1] * (pos + 100.)) + exp(-para[2] * (pos + 100.))) + para[3]);
                         para[0] = par->GetPar2a();
                         para[1] = par->GetPar2b();
                         para[2] = par->GetPar2c();
                         para[3] = par->GetPar2d();
-                        auto q2 = top_tot /
+                        auto q2 = Ctop_tot /
                                   (para[0] * (exp(-para[1] * (pos + 100.)) + exp(-para[2] * (pos + 100.))) + para[3]);
                         q1 = q1 * fTofdQ;
                         q2 = q2 * fTofdQ;
@@ -551,9 +572,12 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
                 }
                 else
                 {
-                    qb = TMath::Sqrt(top_tot * bot_tot);
+                    qb = TMath::Sqrt(Ctop_tot * Cbot_tot);
+                    //qb = fTofdQ*(Ctop_tot + Cbot_tot) / 2.;
                 }
-
+                qb = qb / 200.;
+                if(qb > 70.) qb = qb / 200.;
+               // std::cout<<"qb "<<qb<<std::endl;
                 Double_t parz[3];
                 parz[0] = par->GetPar1za();
                 parz[1] = par->GetPar1zb();
@@ -578,16 +602,25 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
                 LOG(debug) << "y in this event " << pos << " plane " << iPlane << " ibar " << iBar << "\n";
 
                 // Tof with respect LOS detector
-                auto tof = fTimeStitch->GetTime((bot_ns + top_ns) / 2. - header->GetTStart(), "tamex", "vftx");
-                // auto tof_corr = par->GetTofSyncOffset() + par->GetTofSyncSlope() * tof;
-                auto tof_corr = tof - par->GetTofSyncOffset();
+                //auto tof = fTimeStitch->GetTime((bot_ns + top_ns) / 2. , "tamex", "vftx") - fTimeStitch->GetTime(header->GetTStart(), "tamex", "vftx");(bot_ns + par->GetOffset1()) - (top_ns + par->GetOffset2())
+                auto tof = fTimeStitch->GetTime((bot_ns + par->GetOffset1() + top_ns+par->GetOffset2()) / 2.,"tamex","tamex") - fTimeStitch->GetTime(header->GetTStart(), "tamex", "vftx");
+                auto tof_corr = -par->GetTofSyncOffset() + /*par->GetTofSyncSlope() **/ tof;
+               // cout<<"tof_raw "<<tof<<" offset "<<par->GetTofSyncOffset()<<" slope "<<par->GetTofSyncSlope()<<" tof_corr "<<par->GetTofSyncOffset() + par->GetTofSyncSlope() * tof<<endl;
+
+
 
                 // if (parz[1] > 0)
                 // {
                 event.push_back(
-                    { parz[0] + parz[1] * qb + parz[2] * qb * qb, THit, xp, pos, iPlane, iBar, THit_raw, tof_corr });
+                    //{ parz[0] + parz[1] * qb + parz[2] * qb * qb, THit, xp, pos, iPlane, iBar, THit_raw, tof_corr });
+                    { (parz[0] + (parz[1] * qb)/* - 0.153*/ + parz[2] * qb * qb), THit, xp, pos, iPlane, iBar, THit_raw, tof_corr });
+                   // { qb, THit, xp, pos, iPlane, iBar, THit_raw, tof_corr });
                 // }
-
+                //if((parz[0] + (parz[1] * qb) - 0.153 + parz[2] * qb * qb) > 0.){
+                //    cout<<"p1 "<<parz[0]<<" p2 "<<parz[1]<<" p3 "<<parz[2]<<" qb "<<qb<<endl;
+                //    std::cout<<"Q "<<(parz[0] + (parz[1] * qb) - 0.153 + parz[2] * qb * qb)<<std::endl;
+                //}
+               // if(0==parz[0] + (parz[1] * qb) /*- 0.153*/ + parz[2] * qb * qb) exit(0);
                 /* if (parz[0] > 0 && parz[2] > 0)
                  {
                      event.push_back(
@@ -1063,6 +1096,9 @@ void R3BTofDCal2Hit::FinishTask()
     sprint << "n1=" << n1 << " n2=" << n2;
 
     R3BLOG(info, sprint.str());
+
+    cout<<"missing leading edges:\n";
+    cout<<"nev: "<<nev<<" both sides: "<<100.* (double)nev_lmboth / (double)nev<<"%  top: "<<100.* (double)nev_lmt / (double)nev<<"%  bot: "<<100.* (double)nev_lmb / (double)nev<<"%"<<endl;
 }
 
 Double_t R3BTofDCal2Hit::walk(Double_t Q,
--- a/tofd/pars/R3BTofDCal2HitPar.cxx
+++ b/tofd/pars/R3BTofDCal2HitPar.cxx
@@ -1,6 +1,6 @@
 /******************************************************************************
  *   Copyright (C) 2019 GSI Helmholtzzentrum für Schwerionenforschung GmbH    *
- *   Copyright (C) 2019-2023 Members of R3B Collaboration                     *
+ *   Copyright (C) 2019 Members of R3B Collaboration                          *
  *                                                                            *
  *             This software is distributed under the terms of the            *
  *                 GNU General Public Licence (GPL) version 3,                *
@@ -21,7 +21,7 @@
 #include "R3BTofdCalData.h"
 
 #include "FairLogger.h"
-#include "FairRootManager.h"
+#include <FairRootManager.h>
 #include "FairRtdbRun.h"
 #include "FairRunIdGenerator.h"
 #include "FairRuntimeDb.h"
@@ -36,7 +36,9 @@
 #include "TLine.h"
 #include "TMath.h"
 #include "TProfile.h"
+#include "TFile.h"
 #include "TSpectrum.h"
+#include "TSystem.h"
 #include "TVirtualFitter.h"
 
 #include <iostream>
@@ -70,8 +72,11 @@ R3BTofDCal2HitPar::R3BTofDCal2HitPar(const char* name, Int_t iVerbose)
     , fHitPar(NULL)
     , fTofdY(0.)
     , fTofdQ(0.)
-    , fMaxQ(1500)
-    , fNbZPeaks(1)
+	, fMinQ(180)
+    , fMaxQ(300)
+	, fToTMin(190.)
+	, fToTMax(210.)
+    , fNbZPeaks(2)
     , fZfitType("pol1")
     , fTofdTotLow(0.)
     , fTofdTotHigh(200.)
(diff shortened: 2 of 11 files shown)


tofd/calibration/R3BTofDCal2Hit.cxx
```diff
@@ -191,13 +192,13 @@ InitStatus R3BTofDCal2Hit::Init()
...
+    header = (R3BEventHeader*)mgr->GetObject("EventHeader.");
     R3BLOG_IF(fatal, NULL == header, "EventHeader. not found");
 
-    fCalItems = dynamic_cast<TClonesArray*>(mgr->GetObject("TofdCal"));
+    fCalItems = (TClonesArray*)mgr->GetObject("TofdCal");
     R3BLOG_IF(fatal, NULL == fCalItems, "TofdCal not found");
 
-    fCalTriggerItems = dynamic_cast<TClonesArray*>(mgr->GetObject("TofdTriggerCal"));
+    fCalTriggerItems = (TClonesArray*)mgr->GetObject("TofdTriggerCal");
     R3BLOG_IF(fatal, NULL == fCalTriggerItems, "TofdTriggerCal not found");
```
YanzhaoW, line 202: I don't quite understand here. Why do you change dynamic_cast back to C-style cast?

jose-luis-rs: This is something general in the PR, I guess Enis started with these improvements some months ago before the current implementation based on "dynamic_cast".

YanzhaoW: I see. But should the PR be rebased on the HEAD of dev branch?

YanzhaoW:
2000 warnings for a PR with 5000+ lines of changes. Both are too big for a PR. But as I have suggested in the last meeting, if you want to disable the clang-tidy warnings for a code block, you can do:
```cpp
// NOLINTBEGIN

// ... code block ...

// NOLINTEND
``` 
I would still suggest you should fix the warnings as many as possible.

tofd/calibration/R3BTofDCal2Hit.cxx
jose-luis-rs: Yes, rebase the dev branch is the best

tofd/calibration/R3BTofDCal2Hit.cxx
```diff
@@ -434,10 +435,24 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
...
+                    nev++;
+                    if(top_tot > c_range_ns/10 && bot_tot > c_range_ns/10){ nev_lmboth++; both = true;}
+                    if(top_tot > c_range_ns/10){ 
+                        top_tot = bot_tot;
+                        if(!both) nev_lmt++;
+                    }
+                    if(bot_tot > c_range_ns/10){
+                         bot_tot = top_tot;
+                         if(!both) nev_lmb++;
+                    }
```
lbott193, line 454: Does this yield to expected results?
I guess at least charge and y-position will be all over the place for those events.
What do you gain from using those events / how many times does this occur?
Will you use calibration parameters from the correct side later on in the code?

tofd/calibration/R3BTofDCal2Hit.cxx
```diff
@@ -525,8 +545,9 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
                         para[1] = par->GetPolb();
                         para[2] = par->GetPolc();
                         para[3] = par->GetPold();
-                        qb = TMath::Sqrt(top_tot * bot_tot) /
-                             (para[0] + para[1] * pos + para[2] * pow(pos, 2) + para[3] * pow(pos, 3));
+                 //       qb = TMath::Sqrt(Ctop_tot * Cbot_tot) /
+                        qb = TMath::Sqrt(Ctop_tot * Cbot_tot) *
```
lbott193, line 549: Why change this?
This changes calculation of charge completely. Has the parameter calculation also changed?
What about backward compatibility?
Is it necessary to change hit parameters for other users?
It is also different from the other method in lines 560 and 566 which still uses `/`

tofd/calibration/R3BTofDCal2Hit.cxx
```diff
@@ -551,9 +572,12 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
                 }
                 else
                 {
-                    qb = TMath::Sqrt(top_tot * bot_tot);
+                    qb = TMath::Sqrt(Ctop_tot * Cbot_tot);
+                    //qb = fTofdQ*(Ctop_tot + Cbot_tot) / 2.;
                 }
-
+                qb = qb / 200.;
```
lbott193, line 578: Why is the calculated charge divided by 200 or 400?
Shouldn't it be already calibrated by now?

tofd/calibration/R3BTofDCal2Hit.cxx
```diff
@@ -551,9 +572,12 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
...
+                if(qb > 70.) qb = qb / 200.;
```
lbott193, line 579: After you divided by 200 you still expect charges greater than 70 and divide them by 200 again?
Why?

tofd/calibration/R3BTofDCal2Hit.cxx
```diff
@@ -456,13 +473,13 @@ void R3BTofDCal2Hit::Exec(Option_t* option)
...
                 {
-                    auto bot_ns_walk = bot_ns - walk(bot_tot,
+                    auto bot_ns_walk = bot_ns - walk(Cbot_tot,
                                                      par->GetPar1Walk(),
                                                      par->GetPar2Walk(),
                                                      par->GetPar3Walk(),
                                                      par->GetPar4Walk(),
                                                      par->GetPar5Walk());
-                    auto top_ns_walk = top_ns - walk(top_tot,
+                    auto top_ns_walk = top_ns - walk(Ctop_tot,
```
lbott193, line 482: Those variables are not used anywhere else in the code.
Is it known to the user, that walk corrections are not applied?

tofd/calibration/R3BTofDCal2Hit.cxx
lbott193, line 1: I did not have a closer look but at least lines 743 to 845 contain what looks like to be analysis leftovers from when this code was copied from the s494 branch. I think this can be removed. Especially the part with the charge cuts. Not everyone is looking for oxygen, carbon and helium ;)
https://github.com/R3BRootGroup/R3BRoot/pull/842/files#diff-0c573fd7081515e14dae5e88c8a71dcd665a3a49b94c41d56464dad47745ff4dR743-R845

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+		    // if(tof>-153.){
+		    // cout<<"sqrtToT"<<sqrt(top_tot*bot_tot)<<endl; 
+		    //}
+		    //fhlos[iPlane-1]->Fill(iBar,lost);
+		    // if(iPlane==1 && iBar==11){
+		    //fhlost[1][11]->Fill(THit,lost);}
+		    //fhtott[iPlane-1][iBar-1]->Fill(THit,top_tot);
+		    }
+			
+			if(fTofdQ == /*0.0*/ 8. && fParameter == 10){
```
lbott193, line 427: Is `fTofdQ` always expected to be set to 8.?
Or just for your experiment / mistake?
Back in the day this variable was the charge of the calibration beam, but at this point of hit calibration steps no charge was calculated yet, so `fTofdQ==0` was an easy identifier.

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -265,8 +317,11 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
 
             // Shift the cyclic difference window by half a window-length and move it back,
             // this way the trigger time will be at 0.
-            auto top_ns = fTimeStitch->GetTime(top->GetTimeLeading_ns() - top_trig_ns);
-            auto bot_ns = fTimeStitch->GetTime(bot->GetTimeLeading_ns() - bot_trig_ns);
+            auto top_ns =
+                fmod(top->GetTimeLeading_ns() - top_trig_ns + c_range_ns + c_range_ns / 2, c_range_ns) - c_range_ns / 2;
+            auto bot_ns =
+                fmod(bot->GetTimeLeading_ns() - bot_trig_ns + c_range_ns + c_range_ns / 2, c_range_ns) - c_range_ns / 2;
```
lbott193, line 323: This is different than the calculations in TofdCal2Hit.cxx
which uses the previous version with the trig times.
https://github.com/R3BRootGroup/R3BRoot/pull/842/files#diff-0c573fd7081515e14dae5e88c8a71dcd665a3a49b94c41d56464dad47745ff4dR402-R403

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+					fh_tofd_TotPm1[iPlane - 1]->Fill(-iBar - 1, bot_tot);
+				}
+				if(/*top_tot>10. && */top_tot<90.){
+					fh_tofd_TotPm2[iPlane - 1]->Fill(iBar, top_tot);
+					cout<<"Fill Top ToT "<<top_tot<<endl;
+				}
+				if(/*bot_tot>10. && */bot_tot<90.){
+					fh_tofd_TotPm2[iPlane - 1]->Fill(-iBar - 1, bot_tot);
+					cout<<"Fill Bot ToT "<<bot_tot<<endl;
+				}
```
lbott193, line 391: Maybe use some user chosen variables instead of hard coded` 90.`

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+					cout<<"Fill Bot ToT high"<<Cbot_tot<<endl;
+				}
+				if(/*top_tot>10. && */Ctop_tot<70.){
+					fh_tofd_TotPm2[iPlane - 1]->Fill(iBar, Ctop_tot);
+					cout<<"Fill Top ToT low"<<Ctop_tot<<endl;
+				}
+				if(/*bot_tot>10. && */Cbot_tot<70.){
+					fh_tofd_TotPm2[iPlane - 1]->Fill(-iBar - 1, Cbot_tot);
+					cout<<"Fill Bot ToT low"<<Cbot_tot<<endl;
+				}
```
lbott193, line 458: Also here are some hard coded cuts, that might be worth to be changed.

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+				    qb = TMath::Sqrt(Ctop_tot * Cbot_tot);///(parq[0] + parq[1] * posToT + parq[2] * pow(posToT, 2) + parq[3] * pow(posToT, 3));
+				    qbx = TMath::Sqrt(Ctop_tot * Cbot_tot)*(parq[0] + parq[1] * posToT + parq[2] * pow(posToT, 2) + parq[3] * pow(posToT, 3));
+				    qb = qb *
+					    fTofdQ / 200.; // theory says: dE ~ Z^2 but we see quenching -> just use linear and fit the rest
+					qbx = qbx * fTofdQ / 200.;
+				     std::cout<< " hola30 " << qb << " , " << fTofdQ << " , "<<qbx<<std::endl;
+					 if(qbx > 70.){
+						qbx = qbx / 200.;
+						cout<<"XXXXXXX "<< qb << " , " << fTofdQ << " , "<<qbx<<std::endl;
+					 }
```
lbott193, line 696: See my comments in TofdCal2Hit.
I don't understand these changes in charge calculation for the ToT method and division by 200.
Doesn' this produce a lot of output using `cout`?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+							for(int i=0;i<4;i++){
+								cout<<"parf["<<i<<"] "<<parf[i]<<" parq["<<i<<"] "<<parq[i]<<endl;
+							}
+				    auto q2 = Ctop_tot /
+					    (parq[0] * (exp(-parq[1] * (pos + 100.)) + exp(-parq[2] * (pos + 100.))) + parq[3]*1.);
+				    q1 = q1 *
+					    fTofdQ / 200.; // theory says: dE ~ Z^2 but we see quenching -> just use linear and fit the rest
+				    q2 = q2 * fTofdQ / 200.;
+					cout<<"q1 "<<q1<<" q2 "<<q2<<endl;
+				    qb = (q1 + q2) / 2.;
```
lbott193, line 717: See above

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+	    TCanvas* cToTOffset2 = new TCanvas("cToTOffset2", "cToTOffset2", 10, 10, 1000, 900);
+	    cToTOffset2->Divide(1, 2);
+		Double_t fitrange = 25.;
+		if(fParameter == -1) fitrange = 10.;
+	    for (Int_t i = 0; i < fNofPlanes; i++)
+	    {	
+			R3BTofDHitModulePar* par;
+		    for (Int_t j = 0; j < (fPaddlesPerPlane); j++)
+		    {	
+				if(j==45 || j == 46) continue;
```
lbott193, line 1159: why are there 45 or 46 paddles in a plane and why do they have to be skipped?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+
+		fZfitType = "pol2";
+
+	    // TF1* fitz = new TF1("fitz", "[0]*TMath::Power(x,[2])+[1]", min, max);
+	    if (fZfitType != "pol1" && fZfitType != "pol2")
+	    {
+		    R3BLOG(error, "Fit " << fZfitType << " is not allowed, use pol1 or pol2 ");
+		    return -1;
+	    }
+	    auto fitz = new TF1("fitz", fZfitType, min, max);
```
lbott193, line 1517: Interesting.
Did you notice an improvement over the function that I initially used?
Which was something like `[0]*TMath::Power(x,[2])+[1]` if I remember correctly?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+	Int_t rebin2 = 1;
+	Int_t coarse_threshold = 5;
+	Int_t fine_threshold = 2;
+Redo:	
+	Double_t *y;
+		y = (Double_t*)malloc(1000);
+	Double_t *x;
+		x = (Double_t*)malloc(1000);
+	fTofdTotLow = 100.;
+	fTofdTotHigh = 350.;
```
lbott193, line 1747: Why not use the user provided values?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+	}
+	for(int c = 0; c<=n;c++){
+		cout<<"n_fine "<<c<<" x "<<x[c]<<" y "<<y[c]<<endl;
+	}
+	gr2 = new TGraph(n, x, y);
+	gr2->SetTitle("More information;x position in cm;sqrt(q1*q2)");
+	gr2->Draw("A*");
+	f1->DrawCopy("SAME");
+	TF1* f2 = new TF1("f2", "pol3", min, max);
+	f2->SetParameters(para[0], para[1], para[2], para[3]);
```
lbott193, line 1871: Now I'm still unsure why you changed the charge calibration from division to multiplication of this function.
But maybe I missed something in your code.

tofd/pars/R3BTofDCal2HitPar.cxx
lbott193, line 1: Some of your changes broke the github diff.
It is very difficult for me to compare the two versions, so some comments might not apply to your changes but were introduced earlier.

lbott193 (commented):
I only had a quick look at Tofd so far.
Also, I could not check if your changes produce different results than my code.
I will try to get it running and message you then.
>Not advised to use for other experiments yet. Code needs a proper re-work, before it is safe for general use.

If your code should be used for those two experiments only, why not introduce ExpId dependent functions, or do you plan to make it compatible for other experiments?
What are the 
> higher order correction step using physics data

did the number of calibration parameters change, are thy now used in a different way? I might have missed it, because comparing the two versions is not easy due to a lot of changes. Maybe you can show this in one of the next analysis meetings?

I don't know if one should merge if there are hundreds of lines commented and you say the code is not "clean" or it is "running". Why not create a branch and let people who need the code pull from your fork, create a WIP or something.

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
                     // Time differences of one paddle; offset  histo
-                    fhTdiff[iPlane - 1]->Fill(iBar, tdiff);
-
+                    if((sqrt(top_tot*bot_tot)>fToTMin &&sqrt(top_tot*bot_tot)<fToTMax) || ToTexeption){
+		    fhTdiff[iPlane - 1]->Fill(iBar, tdiff);
+		    }
                     auto posToT = TMath::Log(top_tot / bot_tot);
-                    fhSqrtQvsPosToT[iPlane - 1][iBar - 1]->Fill(posToT, sqrt(top_tot * bot_tot));
-
+                   if((sqrt(top_tot*bot_tot)>fToTMin &&sqrt(top_tot*bot_tot)<fToTMax) || ToTexeption){
```
ajedele, line 398: There are 3 identical if statements (if((sqrt(top_tot*bot_tot)>fToTMin &&sqrt(top_tot*bot_tot)<fToTMax) || ToTexeption)). Why?

ToTexeption (spelt incorrectly) is applied, but it's not clear why. When would the ToTexeption condition be added and when not?

If it's always false, why not just remove it?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
-
-                    // Sync of one plane
-                    fhTsync[iPlane - 1]->Fill(iBar, THit);
-                }
-
-                // prepare double exponential fit
-                if (!fTofdSmiley && fTofdQ > 0.1 && fParameter > 1)
-                {
-                    LOG(debug) << "Prepare histo for double exponential fit";
-                    auto par = fHitPar->GetModuleParAt(iPlane, iBar);
```
ajedele, line 427: Why not make !TofdSmiley and fTofdQ>0.1 a consition within (fTofd > 0 && fParameter > 1)?

Why do you need the fTofdQ>0.1 requirement? Would !TofdSmiley be sufficient?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
-
-                // prepare charge fit / quench correction
-                if (fTofdZ == true && fParameter > 1)
-                {
-                    // std::cout<< " hola 20" << std::endl;
-                    LOG(debug) << "Prepare histo for quenching correction";
-                    // get parameter
-                    auto par = fHitPar->GetModuleParAt(iPlane, iBar);
-                    if (!par)
-                    {
```
ajedele, line 453: Redeclaring auto par = fHitPar->GetModuleAt(iPlane, iBar) again. Could be added into 1 (fTofdQ > 0 && fParameter > 1)

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
-                    }
-                }
+                    if((sqrt(top_tot*bot_tot)>fToTMin &&sqrt(top_tot*bot_tot)<fToTMax) || ToTexeption){
+		    fhTsync[iPlane - 1]->Fill(iBar, THit);
+		    }
+		    // Tof with respect LOS detector
+		    auto tof = fTimeStitch->GetTime((top_ns + bot_ns) / 2. - fHeader->GetTStart(), "tamex", "vftx");
+		    if((sqrt(top_tot*bot_tot)>fToTMin &&sqrt(top_tot*bot_tot)<fToTMax) || ToTexeption){
+			    fh1_tofsync[iPlane - 1][iBar - 1]->Fill(tof);
+			    fhTofsync[iPlane-1]->Fill(iBar,tof);
```
ajedele, line 410: Same if statement on line 408 as 403

tofd/pars/R3BTofDCal2HitPar.cxx
ajedele: 8 was the Z of the calibration beam. It's an experiment specific variable. It could probably be changed to fTofdQ > 0 since fTofdQ is set in the macro.

There is no explanation of what fParameter=10 is.

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+				    //fhTot2vsPos[iPlane - 1][iBar - 1]->Fill(posToT, Ctop_tot);
+			    }
+			    // Time of hit
+			    if((sqrt(Ctop_tot*Cbot_tot)>fToTMin && sqrt(Ctop_tot*Cbot_tot)<fToTMax) || ToTexeption){
+				    auto THit = (top_ns + bot_ns) / 2. - par->GetSync();
+				    fhTsync[iPlane - 1]->Fill(iBar, THit);
+			    }
+			    // Tof with respect LOS detector
+			    if((sqrt(Ctop_tot*Cbot_tot)>fToTMin && sqrt(Ctop_tot*Cbot_tot)<fToTMax) /*|| ToTexeption*/){
+				    auto tof = fTimeStitch->GetTime((top_ns + bot_ns) / 2. - fHeader->GetTStart(), "tamex", "vftx");
```
ajedele, line 525: Again, repeated if statement...

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+						fhTofsync[iPlane-1]->Fill(iBar,tof);
+				    	fh1_tofsync[iPlane - 1][iBar - 1]->Fill(tof);	
+					}
+					else{
+						fhTofsync[iPlane-1]->Fill(iBar,tof - par->GetTofSyncOffset());
+				    	fh1_tofsync[iPlane - 1][iBar - 1]->Fill(tof - par->GetTofSyncOffset());
+					}
+			    }
+			}
+		    if(fParameter==0){
```
ajedele, line 536: What is fParameter=0? Does this need to be included?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+
+			    if((sqrt(Ctop_tot*Cbot_tot)>fToTMin && sqrt(Ctop_tot*Cbot_tot)<fToTMax) || ToTexeption){
+				    fh1_walk_t[iPlane-1][iBar-1]->Fill(Ctop_tot,top_ns);
```
ajedele, line 539: 😤

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+				    fh1_walk_b[iPlane-1][iBar-1]->Fill(Cbot_tot,bot_ns);
+			    }
+		    }
+
+
+		    else if ((fTofdQ > 0 && fParameter > 1 && fParameter != 10) || fParameter == 3)
+		    {
```
ajedele, line 546: Is this condition every entered?

The initial if statement is (fTofd>0 && fParameter>1 && fParameter!=10). fParameter==3 is greater than 1 and not 10

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+			    if((sqrt(Ctop_tot*Cbot_tot)>fToTMin && sqrt(Ctop_tot*Cbot_tot)<fToTMax) || ToTexeption){
+				    fhTdiff[iPlane - 1]->Fill(iBar, tdiff);
+			    }
+			    // Time of hit
+			    auto THit = (bot_ns + top_ns) / 2. - para->GetSync();
+			    if((sqrt(Ctop_tot*Cbot_tot)>fToTMin && sqrt(Ctop_tot*Cbot_tot)<fToTMax) || ToTexeption){
+			    }
+			    // Sync of one plane
+			    if((sqrt(Ctop_tot*Cbot_tot)>fToTMin && sqrt(Ctop_tot*Cbot_tot)<fToTMax) || ToTexeption){
+				    fhTsync[iPlane - 1]->Fill(iBar, THit);
```
ajedele, line 609: 😤

tofd/pars/R3BTofDCal2HitPar.cxx
ajedele: 200 is the ToT value consistent with where the beam was. It's a hard-coded value - probably not a good ideal

The code effectively says take the sqrt(ToT1*ToT2) (should be approx. 200), multiply it by 8 (charge of the O-beam at ToT=200) and dividing by the value it should be. Effectively, this is a charge calibration at 1 point.

Agree, it should not be hard-coded, but set as a variable.

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+				
+				if (multihits[iPlane - 1][iBar - 1] < 2 /*&& qb>190. */ || 1)
+			    {
+				    fhQvsPos[iPlane - 1][iBar - 1]->Fill(posToT, qb);
+					fhQXvsPos[iPlane - 1][iBar - 1]->Fill(posToT, qbx);
+			    }
+	
+			}
+	/*	    if(fParameter==5)
+		    {
```
ajedele, line 740: Is this code needed? If not, just delete

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+		}
+		else
+		{
+			++bot_i;
+		}
+	    }
+	}
+    }
+	void R3BTofDCal2HitPar::SetHistogramsNULL(Int_t iPlane, Int_t iBar){
+		fhTdiff[iPlane - 1] = NULL;
```
ajedele, line 789: This could be done in the header using in the syntax:

TH1F *fhNameofHisto[NofPlanes][NofBars]{};

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+
+	    if (NULL == fhTdiff[iPlane - 1])
+	    {
+		    char *strName1;
+			strName1 = (char*)malloc(255);
+		    char *strName2;
+			strName2 = (char*)malloc(255);
+		    sprintf(strName1, "Time_Diff_Plane_%d", iPlane);
+		    sprintf(strName2, "Time Diff Plane %d", iPlane);
+		    fhTdiff[iPlane - 1] = new TH2F(strName1, strName2, 50, 0, 50, 4000, -20., 20.);
```
ajedele, line 820: Could use a string instead. Would save 4 lines of code per Histogram declaration

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+		    sprintf(strName2, "QX vs Pos Plane %d Bar %d", iPlane, iBar);
+		    //cout<<"sprint7 "<<strName<<" "<<strName2<<endl;
+			fhQXvsPos[iPlane - 1][iBar - 1] = new TH2F(strName, strName2, 2000, -100, 100, /*max_charge , 0., 2.*max_charge*/ 1600, 0., 16.);
+		    fhQXvsPos[iPlane - 1][iBar - 1]->GetYaxis()->SetTitle("Charge");
+		    fhQXvsPos[iPlane - 1][iBar - 1]->GetXaxis()->SetTitle("Position in cm");
+			free(strName);
+			free(strName2);
+	    }
+	   /* if (NULL == fhZoffset[iPlane - 1][iBar - 1])
+	    {
```
ajedele, line 1005: Are these needed? If not, please delete

tofd/pars/R3BTofDCal2HitPar.cxx
ajedele: I assume this was a relic from a previous version or a sanity check. It should be removed

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+				Double_t a_top = 1.;	
+				Double_t b_bot = 0.;	
+				Double_t b_top = 0.;
+				Double_t Max1 = 1.;
+				Double_t Max2 = 1.;
+				Double_t Max3 = 0.;
+				Double_t Max4 = 0.;
+
+			    if(j<44) par = fHitPar->GetModuleParAt(i + 1, 1 + j);
+			    else par = fHitPar->GetModuleParAt(i + 1, j - fPaddlesPerPlane - 2);
```
ajedele, line 1175: When would you need the else statement?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+			    fhSqrtQvsPosToT[i][j]->Write();
+				}
+			    LOG(warning) << " Plane  " << i + 1 << " Bar " << j + 1 << " ToT Offset  " << offset << "\n";
+				LOG(info) <<"TOP " << 1. / sqrt(exp(offset)) << " BOT " <<sqrt(exp(offset))<< "\n";
+			    par->SetToTOffset1(sqrt(exp(offset)));
+			    par->SetToTOffset2(1. / sqrt(exp(offset)));
+				
+
+/*				if (fh_tofd_TotPm1[i])
+			    {
```
ajedele, line 1207: Are these lines of code needed? If not, please delete

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+				    Int_t binmax = histo_py->GetMaximumBin();
+				    max = histo_py->GetXaxis()->GetBinCenter(binmax);
+				    Double_t maxEntry = histo_py->GetBinContent(binmax);
+				    auto* fgaus = new TF1("fgaus", "gaus(0)", max - 0.3, max + 0.3); /// TODO: find best range
+				    fgaus->SetParameters(maxEntry, max, 20);
+				    histo_py->Fit("fgaus", "QR0");
+				    Double_t offset1 = par->GetOffset1();
+				    Double_t offset2 = par->GetOffset2();
+				    max = fgaus->GetParameter(1) + offset1 - offset2; /// TODO: needs to be tested
+				    // max = max+offset1-offset2;
```
ajedele, line 1589: Offset is added/subtracted here again. I don't think this is correct.

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+			    LOG(info) << " Plane  " << i + 1 << " Bar " << j + 1 << " ToT Offset  " << offset << " Lambda " << lambda
+				    << "\n";
+			    par->SetLambda(lambda);
+		    }
+	    }
+	    fHitPar->setChanged();
     }
-    if (NULL == fhTot2vsPos[iPlane - 1][iBar - 1] && !fTofdSmiley)
+    /*  void R3BTofDCal2HitPar::calcwalk(Double_t totLow, Double_t totHigh)
+	{
```
ajedele, line 1644: Needed?

tofd/pars/R3BTofDCal2HitPar.cxx
ajedele: Shouldn't these be set in the macro? Depending on the experiment, the range is different

tofd/pars/R3BTofDCal2HitPar.cxx
ajedele: Enis looked into this. When he flipped the sign, the data converged. Before that, it diverged

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
-    {
-        // Determine effective speed of light in [cm/s] for each paddle
-        LOG(info) << "Calling function";
-        calcVeff();
-        // Determine light attenuation lambda for each paddle
-        LOG(info) << "Calling function calcLambda";
-        calcLambda(fTofdTotLow, fTofdTotHigh);
-    }
-    else if (fParameter == 3)
+/*   void R3BTofDCal2HitPar::calcZoffset()
```
ajedele, line 2003: This function is commented out. Is it needed?

tofd/pars/R3BTofDCal2HitPar.cxx
lbott193: Ok I agree for `qb` here. If there are no charge calibration parameters this is a good first guess.

But isn't the function that gives this parq parameters in line 688 fitted with just `sqrt(ToT1*ToT2)` as y-values filled in lines 514 or 596? With a pol3 fit this should describe the smiley/frowny/whatever, at the ToT that was aimed for in the calibration beam?

In the classical case I expect `sqrt(ToT1*ToT2)` to rise on the ends of a paddle. 
If I muliply the `sqrt(ToT1*ToT2)` (say 210) at this given `posToT` with the value of the function at this point (ideal case also 210 since we only have calibration beam and no fragments) and 8, i end up with something large. That I divide by 200 which is still large (`>70.`) and because of that I divide again by 200 which is then in the range were i want it to be but not 8 (210\*210\*8/200/200=8.82) and even further away from 8 than the qb method you described (210*8/200=8.4) which doesn't use the fit function??

Before, as commented in line 687 we would've had with this example: 210/210*8=8 were the second 210 is the value of the function.

But maybe I'm just slow today and missed something in the code or the function definition changed...

Anyway, I also don't see the benefit of remembering what value the ToT was set to in the calibration beam each time one wants to run the Cal2Hit macro, why not use the info from the fit and in the macro only the charge?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
-                fhTot1vsTot2[i][j]->Write(); // control histogram ToT Pm1 vs ToT Pm2
-            */
-        }
-    }
+void R3BTofDCal2HitPar::FinishTask()
+{	
+
+	if (fParameter == 10)
+	{
+		// Determine time offset of the 2 PMTs of one paddle. This procedure
```
ajedele, line 2054: It is only difference between fParameter==10 and fParameter==1 the difference in time between the up/down PMT? Does this need a separate step?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+		LOG(info) << "Calling function calcToTOffset";
+		calcToTOffset(fTofdTotLow, fTofdTotHigh);
+		// Determine sync offset between paddles
+
+		for (Int_t i = 0; i < fNofPlanes; i++)
+		{
+		fhTofsync[i]->Write();
+		fh_tofd_TotPm[i][0]->Write();
+		fh_tofd_TotPm[i][1]->Write();
+		fh_tofd_TotPm[i][2]->Write();
```
ajedele, line 2068: Is fh_tofd_TotPm[i][2] used in the calibrations? If not, remove it?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+				fh1_tofsync[i][j]->Write();
+			}
+		fhTofsync[i]->Write();
+		fh_tofd_TotPm[i][0]->Write();
+		fh_tofd_TotPm[i][1]->Write();
+		fh_tofd_TotPm[i][2]->Write();
+		}*/
+	}
+	else if(fParameter == 0)
+	{
```
ajedele, line 2105: See previous fParameter==0 comment

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+	}
+	else if (fParameter == 2)
+	{
+		//Calculate ToF
+		LOG(info) << "Calculate ToF";
+		for (Int_t i = 0; i < fNofPlanes; i++)
+		{
+			for (Int_t j = 0; j < fPaddlesPerPlane; j++)
+			{
+				auto par = fHitPar->GetModuleParAt(i + 1, j + 1);
```
ajedele, line 2120: Where does the par get used in this loop?

tofd/pars/R3BTofDCal2HitPar.cxx
```diff
@@ -297,1099 +352,2668 @@ void R3BTofDCal2HitPar::Exec(Option_t* option)
...
+		}
+	}
+	else if (fParameter == 3)
+	{
+		// calculation of position dependend charge
+		if (fTofdSmiley)
+		{
+			
+		//	LOG(info) << "Calling function smiley";
+		//	Double_t para2[4];
```
ajedele, line 2157: See previous comments about commented code

ajedele: Can change para2[4]; for (int i=0; i<4;i++) to para2[4]{};

tofd/pars/R3BTofDCal2HitPar.cxx
lbott193: Interesting. Is there a mixup somewhere that causes this issue? I would expected something like I described in https://github.com/R3BRootGroup/R3BRoot/pull/842#discussion_r1164626320

enlorenz:
> I only had a quick look at Tofd so far. Also, I could not check if your changes produce different results than my code. I will try to get it running and message you then.
> 
> > Not advised to use for other experiments yet. Code needs a proper re-work, before it is safe for general use.
> 
> If your code should be used for those two experiments only, why not introduce ExpId dependent functions, or do you plan to make it compatible for other experiments? What are the
> 
The code should be of course compatible with other experiments and especially for future experiments. Backwards compatibility is another question. I would like to avoid ExpID dependent functions, since most of the dependencies should be possible to have flags and/or variables set in the macro. Individual functions make the code unnecessarily blown up. If it is OK for everyone, I would actually like to make this code a WiP version, so that it can be merged in the future.
 
> > higher order correction step using physics data
> 
> did the number of calibration parameters change, are thy now used in a different way? I might have missed it, because comparing the two versions is not easy due to a lot of changes. Maybe you can show this in one of the next analysis meetings?
>
Yes, I will definitely present in detail what it is. Briefly, it is an extra calibration step which can or cannot be used, depending on the situation. This adjusts the values of the final charge parameters, so that the application of the parameters in the Cal2Hit itself does not change. 

> I don't know if one should merge if there are hundreds of lines commented and you say the code is not "clean" or it is "running". Why not create a branch and let people who need the code pull from your fork, create a WIP or something.

Yes, that should be the best option for now.

tofd/calibration/R3BTofDCal2Hit.cxx
enlorenz: Good point, I need to re-check if this still holds up. This piece of code is (at least) a relic of the investigation of missed leading edges, which lead to unreasonably high ToTs. 1/10th of the range is an arbitrary threshold, but it is a clear indicator if the ToT of one side is way too high. Setting it to the same value as the other pmt at least allows for some recycling of said event.

The occurrences are low though: ~0.092% topside, ~0.094% bottomside and ~0.028% both sides.

tofd/calibration/R3BTofDCal2Hit.cxx
enlorenz: As Andrea mentioned further below, using  /  did not yield correct results. Using the multiplication on the other hand made the channels converge. As I did not change the shape of the parameters themselves, it should be backwards compatible still in this case.
I did not further explore the double exponential, therefore I don't know if there it should be a  /  or  *

tofd/calibration/R3BTofDCal2Hit.cxx
enlorenz: This is one of the points that are (for now) hard coded and need to be resolved. The Charge alignment is run over scaled qb, so this needs to be scaled to the same level. 200 is arbitrary and is roughly the mean ToT value of Oxygen in this calibration.

tofd/calibration/R3BTofDCal2Hit.cxx
enlorenz: It appeared that sometimes the ToTs carry another factor of said scaling factor. Where this comes from is not clear at the moment, but dividing those events again by 200 solves the issue, 70 is chosen randomly as it just says "sufficiently high". But these two lines should just be considered a "hot fix".

tofd/calibration/R3BTofDCal2Hit.cxx
enlorenz: This fact should be mentioned clearly, I agree.

tofd/pars/R3BTofDCal2HitPar.cxx
enlorenz: fParameter == 10 is a relic and can be removed.

fTofdQ is a flag if it is set to any value > 0 in the macro, it's basically a boolian statement at this point and should be treated as such.

tofd/pars/R3BTofDCal2HitPar.cxx
enlorenz: This code is obsolete.

tofd/pars/R3BTofDCal2HitPar.cxx
enlorenz: same here. The fh_tofd_TotPm[iPlane - 1][0] is only for checking values. Agree that the threshold should be easier to handle.

tofd/pars/R3BTofDCal2HitPar.cxx
enlorenz: Interestingly so. though calling "pol2" and [0]*TMath::Power(x,[2])+[1] should yield the same result, the latter for some reason did not fit consistently....

tofd/pars/R3BTofDCal2HitPar.cxx
enlorenz: Yes they should. For some reason I needed to make sure at this point that these values are used.

tofd/pars/R3BTofDCal2HitPar.cxx
lbott193: Ok, if it improves the fit it is of course better. But I would not expect same results. The function I used was an observation that the light quenching is almost linear in the ToFD case and not something quadratic or other models like Birk or so, as you can see in the Tofd paper https://repository.gsi.de/record/250622 figure 16. Some paddles saw more quenching others less, that's why I did fit the exponent itself.

tofd/calibration/R3BTofDCal2Hit.cxx
lbott193: Is there a plan to fix this or are walk corrections done in other analysis steps like tracking?

tofd/pars/R3BTofDCal2HitPar.cxx
lbott193: Since you sent me your analysis report I think i finally understand why this is done with pol2 and pol1.
I think you've done some steps that are at best not necessary and ultimatly came to the same point I mentioned in the previous comment, but maybe we can discuss this somewhere else.

ajedele:
General comment/question (and maybe @lbott193 you can help): There are 2 ways to do the position dependent charge correction - smiley and double exponential. In theory, both should give the same results. Both are also present in the calibration.

However, the double exponential currently crashes because the parameters are not included in the R3BTofdHitPar or R3BTofdHitModulePar. The smiley, from my understand is also simpler and produces better results, faster. 

Should we continue to support the double exponential if it's currently not possible to use?

tofd/pars/R3BTofDCal2HitPar.h
```diff
@@ -193,6 +198,7 @@ class R3BTofDCal2HitPar : public FairTask
      */
     Double_t walk(Double_t Q, Double_t par1, Double_t par2, Double_t par3, Double_t par4, Double_t par5);
 
+
     /**
      * Method for calculation of saturation.
```
ajedele, line 203: I do not believe the saturation function is used. Are there plans to implement this or not? @lbott193

lbott193:
> General comment/question (and maybe @lbott193 you can help): There are 2 ways to do the position dependent charge correction - smiley and double exponential. In theory, both should give the same results. Both are also present in the calibration.
> 
> However, the double exponential currently crashes because the parameters are not included in the R3BTofdHitPar or R3BTofdHitModulePar. The smiley, from my understand is also simpler and produces better results, faster.

I can see them in the current R3BTofdHitModulePar.h they are named (maybe one should come up with more descriptive names):
fPar1a-c and fPar2a-c: https://github.com/R3BRootGroup/R3BRoot/blob/dev/tofd/pars/R3BTofDHitModulePar.h#L154-L155

Why it doesn't work anymore, I don't know.

> 
> Should we continue to support the double exponential if it's currently not possible to use?

Charge correction with tdfiff and double exponential was difficult in the past under certain circumstances, for example interference effects on the electronics card. The ToT method is less sensitive to such effects and at that time we got better results. You might want to check that for your setup and use the best option, that's why I left both methods in.

tofd/pars/R3BTofDCal2HitPar.h
lbott193: I think TofdCal2Hit was introduced in PR https://github.com/R3BRootGroup/R3BRoot/pull/90
There you can see that the saturation function is commented:
https://github.com/michael-heil/R3BRoot/blob/7f4b0d57dc2a9f2691dfdd7020a7da6a0553b967/tof/R3BTofdCal2Hit.cxx#L575-L577

I don't know how to track file changes on this webpage but I never used the saturation function, maybe @michael-heil can answer your question.

tofd/calibration/R3BTofDCal2Hit.cxx
ajedele: In this case, I agree with Lukas. Multiplying is incorrect

enlorenz: "In this case" meaning for the double exponential? Or for the "smiley function" too?

tofd/pars/R3BTofDCal2HitPar.cxx
enlorenz: The division is applied only after the smiley corrected value of qb is calculated, so it should not diverge on the edges of the bar anymore. But maybe I now am missing something myself... Maybe this is best further discussed in person.:)

tofd/pars/R3BTofDCal2HitPar.cxx
enlorenz: Seven lines below as: par->SetTofSyncOffset(tof_offset - fMeanTof);

tofd/pars/R3BTofDCal2HitPar.cxx
enlorenz: Have split step "3" into two, the commented lines are in step "-3". For now leave it to be ignored. Can be removed, once the split is accepted.

PR #849, opened 2023-04-21
New propagator that includes energy loss in air

ManuelXarepe:
New function to include energy loss in the R3BTPropagator class. This new function is necessary for the tracking of the 40Ar beam used for the sweep run of the RPC. This function also allows for better control of the step used to propagate the particle, due to this the option to run without energy loss was implemented. 

As it is right now it is only possible to calculate the energy loss in the air but can be improved in future implementations if needed. 

---

Checklist:

### DIFF
--- a/tracking/R3BTPropagator.cxx
+++ b/tracking/R3BTPropagator.cxx
@@ -346,6 +346,57 @@ Bool_t R3BTPropagator::PropagateToPlaneRK(R3BTrackingParticle* particle,
     return kTRUE;
 }
 
+Bool_t R3BTPropagator::PropagateToPlaneRK_eloss(R3BTrackingParticle* particle,
+                                                const TVector3& point1,
+                                                const TVector3& point2,
+                                                const TVector3& point3,
+                                                double& Tot_Tof,
+                                                double step,
+                                                bool eloss)
+{
+
+    std::array<double, N_PARTICLE_INFO> vecRKIn{};
+    std::array<double, N_PARTICLE_INFO> vecOut{};
+
+    particle->GetPosition(vecRKIn.data());
+    particle->GetCosines(&vecRKIn[3]);
+
+    const TVector3 norm = ((point2 - point1).Cross(point3 - point1)).Unit();
+    TVector3 intersect;
+
+    LineIntersectPlane(particle->GetPosition(), particle->GetMomentum(), point1, norm, intersect);
+    Double_t length = 0.;
+    bool continue_while = true;
+    if ((intersect - particle->GetPosition()).Mag() - step < step)
+    {
+        LOG(error) << "step " << step << " is too big compared with wanted step of "
+                   << (intersect - particle->GetPosition()).Mag();
+        return kFALSE;
+    }
+    // propagates the particle with a fixed step
+    while (LineIntersectPlane(particle->GetPosition(), particle->GetMomentum(), point1, norm, intersect) &&
+           continue_while)
+    {
+        if ((intersect - particle->GetPosition()).Mag() - step < step)
+        {
+            step = (intersect - particle->GetPosition()).Mag();
+            continue_while = false;
+        }
+        length = fFairProp->OneStepRungeKutta(particle->GetCharge(), step, vecRKIn.data(), vecOut.data());
+        // update particle info
+        update_particle(particle, length, vecOut, eloss);
+        // update position and momentum array
+        std::copy_n(vecOut.begin(), vecOut.size(), vecRKIn.begin());
+        if (particle->GetBeta() < 0)
+        {
+            LOG(error) << "particle was stopped did not reach destination";
+            return kFALSE;
+        }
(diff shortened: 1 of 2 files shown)


jose-luis-rs:
Please, rebase the main dev branch to include the clang-format check

ManuelXarepe:
After a great battle with clang-tidy and format, I think it is ok now! Open for comments.

tracking/R3BTPropagator.cxx
```diff
@@ -370,4 +424,50 @@ Bool_t R3BTPropagator::LineIntersectPlane(const TVector3& pos,
...
+    {
+        particle->SetPosition(vecOut.data());
+        particle->SetCosines(&vecOut[3]);
+        particle->AddStep(length);
+    }
+}
+
+Double_t R3BTPropagator::Energy_loss_in_air(double beta, double step, double charge, double mass)
+{
+    const double me = 0.51099895; // NOLINT // MeV/c^2
```
YanzhaoW: What is the meaning of 'me'?

tracking/R3BTPropagator.cxx
```diff
@@ -370,4 +424,50 @@ Bool_t R3BTPropagator::LineIntersectPlane(const TVector3& pos,
...
+
+    // Values taken from https://pdg.lbl.gov/2022/AtomicNuclearProperties/HTML/air_dry_1_atm.html
+    const double ZoAair = 0.49919;
+    const double I = 85.7e-6;    // NOLINT // MeV (mean excitation energy) taken from PDG site
+    const double rho = 0.001205; // g/cm^3 (density of air at STP)
+    const double K = 0.307075;   // NOLINT // MeV mol^-1 cm^2 (constant factor)
+    const double gamma = 1 / (sqrt(1 - pow(beta, 2)));
+    const double Tmax =
```
YanzhaoW: I think we should definitely have better naming here. For "ZoAair", "I", "K" and "Tmax", I have no idea what they mean. Of course, I could go to the literature to look them up. But with better naming, it could same me and other people a lot of time to understand what's going here.

YanzhaoW (commented):
Some comments about the naming. And many of the const variables can be declared as `constexpr`. It would improve the speed of algorithm a lot.

YanzhaoW:
Hi, 

Thanks for your contribution. I have never used this class before. But from its location, I guess it's not dedicated to a specific detector. Could you explain very shortly what this TPropagator class does?

tracking/R3BTPropagator.cxx
jose-luis-rs: It seems the electron mass

ManuelXarepe: I will turn some of these variables to contexpr, no problem, and yes is the mass of the electron.

tracking/R3BTPropagator.cxx
ManuelXarepe: Hi, changing the name of some of these variables, in my opinion, will make the code less readable for those that are used to look at the Bethe Bloch equation. For some of these variables I already wrote the meaning of them, as is the case for I, the mean excitation energy and K, a constant factor. For others like the Z over A,  "me", and Tmax I assumed that they were self-explanatory, with the context where they are inserted.

tracking/R3BTPropagator.cxx
YanzhaoW: So in this case, would "electron_mass" be a better name?

tracking/R3BTPropagator.cxx
YanzhaoW: "I" usually means current value in a circuit or the intensity in nuclear physics. Maybe "mean_Excit_E"? K could also mean temperature. Here maybe "const_k"? Generally I don't recommend to use single (or two) letter for the name of a variable. They could mean so many other things. 

And if "ZoAair" means "Z over A" value for air, better just write them out like Z_over_A in my opinion.

ManuelXarepe: As I said before, in the context where they are introduced, Ithink they are explicit as the variables are defined as you would find them in many of the Bethe Bloch equations. But in the end of the day, they are just variable names, I will not fight to keep something that is a 5 minutes fix. Iwill change them to be more readable to anyone. Unfortunately today I lack time to implement and check changes due to other problems that appeared due to lustre being down. Tomorrow or later today I will update the PR.

YanzhaoW: No problem. But yes, they are very minor issues.

jose-luis-rs:
Hi @ManuelXarepe 
Is this PR ready for the merger?
Cheers

ManuelXarepe:
Hello, @jose-luis-rs
Tomorrow I will rebase some new changes. I had a look at the code and decided it could be improved, sorry for not doing it quicker.
Cheers

ManuelXarepe:
Sorry for the delay. Should be completed now.

PR #857, opened 2023-05-06
clang-format-15 for all remaining .C .cxx .h .hh

inkdot7:
Changes needed due to #828 for other merge requests to pass CI format checks.

I have *not* reviewed them carefully, just slightly with `git diff --word-diff`.  I believe review is the responsibility proponents of clang-format-15.

PR #858, opened 2023-05-06
Remove all layout init

inkdot7:
Let's see if the code compiles without those _LAYOUT_INIT macros.  If so, then please consider merge this mass-removal of cruft.


inkdot7:
#859 is better.  This is based on something (#857) which likely turns well-formatted code into something not digestible.

PR #866, opened 2023-06-08
TwiM drift time calib fixed, suppress unnecessary cout

ryotani:
The drift time calibration for TwiM in s467/s444 (2020) has to be independent of the standard configuration since two MDPP16 were used for section 0 for those cases. The mapping of the time reference was wrong previously, and the R3BCoarseTimeStitch has been removed since it was causing fake reconstructions. The code has been simplified.

Describe your proposal.

---

Checklist:

### DIFF
--- a/twim/calibration/R3BTwimMapped2Cal.cxx
+++ b/twim/calibration/R3BTwimMapped2Cal.cxx
@@ -28,13 +28,15 @@
 #include "FairRuntimeDb.h"
 
 // TWIM headers
-#include "R3BCoarseTimeStitch.h"
 #include "R3BEventHeader.h"
 #include "R3BLogger.h"
 #include "R3BTwimCalData.h"
 #include "R3BTwimCalPar.h"
 #include "R3BTwimMapped2Cal.h"
 #include "R3BTwimMappedData.h"
+
+constexpr int S444 = 444;
+constexpr int S467 = 467;
 
 // R3BTwimMapped2Cal::Default Constructor --------------------------
 R3BTwimMapped2Cal::R3BTwimMapped2Cal()
@@ -60,13 +62,9 @@ R3BTwimMapped2Cal::R3BTwimMapped2Cal(const TString& name, Int_t iVerbose)
     , fExpId(0)
     , fOnline(kFALSE)
 {
-    CalEParams.resize(fNumSec);
-    PosParams.resize(fNumSec);
-    for (Int_t s = 0; s < fNumSec; s++)
-    {
-        CalEParams[s] = NULL;
-        PosParams[s] = NULL;
-    }
+    CalEParams.resize(fNumSec, nullptr);
+    PosParams.resize(fNumSec, nullptr);
+    fTwimCal.resize(fNumSec);
 }
 
 void R3BTwimMapped2Cal::SetParContainers()
@@ -127,13 +125,18 @@ void R3BTwimMapped2Cal::SetParameter()
     }
 
     // Count the number of dead anodes or not used
-    for (Int_t s = 0; s < fNumSec; s++)
+    for (Int_t secId = 0; secId < fNumSec; secId++)
     {
         Int_t numdeadanodes = 0;
-        for (Int_t i = 0; i < fNumAnodes; i++)
-            if (CalEParams[s]->GetAt(fNumEParams * i + 1) == -1 || fCal_Par->GetInUse(s + 1, i + 1) == 0)
+        for (Int_t anodeId = 0; anodeId < fNumAnodes; anodeId++)
+        {
+            if (CalEParams[secId]->GetAt(fNumEParams * anodeId + 1) == -1 ||
+                fCal_Par->GetInUse(secId + 1, anodeId + 1) == 0)
+            {
                 numdeadanodes++;
-        R3BLOG(info, "Dead (or not used) anodes in section " << s + 1 << ": " << numdeadanodes);
+            }
+        }
+        R3BLOG(info, "Dead (or not used) anodes in section " << secId + 1 << ": " << numdeadanodes);
     }
     return;
 }
@@ -164,9 +167,6 @@ InitStatus R3BTwimMapped2Cal::Init()
     frm->Register("TwimCalData", "TWIM_Cal", fTwimCalDataCA, !fOnline);
     Reset();
 
-    // Definition of a time stich object to correlate VFTX times
-    fTimeStitch = new R3BCoarseTimeStitch();
-
     SetParameter();
     return kSUCCESS;
 }
@@ -179,8 +179,44 @@ InitStatus R3BTwimMapped2Cal::ReInit()
     return kSUCCESS;
 }
 
+void R3BTwimMapped2Cal::CalAnode::Init()
+{
+    fmult = 0;
+    fE.clear();
+    fDT.clear();
+    fE.resize(MAXNUMANODE, NAN);
+    fDT.resize(MAXNUMANODE, 0);
+}
+
+R3BTwimMapped2Cal::CalSection::CalSection()
+{
+    fAnode.clear();
+    fAnode.resize(MAXNUMANODE);
+
+    fTref.clear();
+    fTref.resize(MAXNUMTREF);
+
+    fTrig.clear();
+    fTrig.resize(MAXNUMTRIG);
+}
+void R3BTwimMapped2Cal::CalSection::Init()
+{
+    for (auto& val : fAnode)
+    {
+        val.Init();
+    }
+    for (auto& val : fTref)
+    {
+        val.Init();
+    }
+    for (auto& val : fTrig)
+    {
+        val.Init();
+    }
+}
+
 // -----   Public method Execution   --------------------------------------------
-void R3BTwimMapped2Cal::Exec(Option_t*)
+void R3BTwimMapped2Cal::Exec(Option_t* /*option*/) // NOLINT(readability-function-cognitive-complexity)
 {
     // Reset entries in output arrays, local arrays
     Reset();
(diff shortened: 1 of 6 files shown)


ryotani:
Hi, I need your help @YanzhaoW @klenze. This is way beyond my expertise in writing C++ code. Initially, I wanted to change the channel of Time reference for the drift time correction, resulting in me having to almost rewrite the code since CI cannot be happy with any modifications.
It's not my code originally, but previous authors were using fE[section][mult][ChId] and a similar multi-dimensional array fDT[][][]. I tried to define proper classes as CalSection, CalAnode and seems working as expected. But I cannot find the solution to answer the message as:

> assigning newly created 'gsl::owner<>' to non-owner 'std::array<R3BTwimMapped2Cal::CalSection *, 4>::value_type' (aka 'R3BTwimMapped2Cal::CalSection *') [cppcoreguidelines-owning-memory]

Could you advise me to solve this?
Other comments might be able to be accessed by me.. But I would say this clang-tidy creates messes for people as they want to fix only a few lines, since clang-format will change many lines then clang-tidy will start pointing out issues caused not by the latest contributors....

YanzhaoW:
Hi, @ryotani 

No problem. The warning (yes, the warning message is not clear) is caused by the usage of `new` operator. Instead of using new operator, you should create the resource by using std::make_unique<ClassType>(). By this way, you can get rid of unreleased memory or deleting a pointer twice. Google unique pointer of C++ or search it in youtube and you could get a very good understanding of why we need it.

But I will check your code and give some reviews tomorrow. If you have more questions, we could discuss tomorrow.

twim/calibration/R3BTwimMapped2Cal.h
```diff
@@ -88,13 +94,51 @@ class R3BTwimMapped2Cal : public FairTask
...
+        CalSection();
+        void Init();
+        auto GetAnode(Int_t id) { return fAnode[id]; };
+        auto GetTrig(Int_t id = 0) { return fTrig[id]; };
+        auto GetTref(Int_t id = 0) { return fTref[id]; };
+
+      private:
+        std::array<CalAnode*, MAXNUMCH> fAnode;
+        std::array<CalAnode*, MAXNUMTREF> fTref;
+        std::array<CalAnode*, MAXNUMTRIG> fTrig;
```
YanzhaoW: These can be changed into `std::vector<CalAnode> fAnode`, etc

twim/calibration/R3BTwimMapped2Cal.cxx
```diff
@@ -179,8 +183,49 @@ InitStatus R3BTwimMapped2Cal::ReInit()
     return kSUCCESS;
 }
 
+void R3BTwimMapped2Cal::CalAnode::Init()
+{
+    fmult = 0;
+    for (Int_t mult = 0; mult < MAXMULT; mult++)
+    {
+        fE[mult] = NAN;
```
YanzhaoW: If you use `std::vector`, you could simply use `resize` member function to zero initialise a certain number of elements:
```cpp
fE.clear();
fE.resize(mult);
```

twim/calibration/R3BTwimMapped2Cal.cxx
```diff
@@ -179,8 +183,49 @@ InitStatus R3BTwimMapped2Cal::ReInit()
...
+        fDT[mult] = 0;
+    }
+}
+
+R3BTwimMapped2Cal::CalSection::CalSection()
+{
+    for (int ChId = 0; ChId < MAXNUMCH; ChId++)
+    {
+        fAnode[ChId] = new CalAnode();
```
YanzhaoW: Here with std::vector (without for loop):
```cpp
fAnode.clear();
fAnode.resize(MAXNUMCH);
```

twim/calibration/R3BTwimMapped2Cal.cxx
```diff
@@ -191,97 +236,86 @@ void R3BTwimMapped2Cal::Exec(Option_t*)
...
 
-    for (Int_t s = 0; s < fNumSec; s++)
-        for (Int_t i = 0; i < (fNumAnodes + fNumAnodesRef + fNumAnodesTrig); i++)
+    for (Int_t sec = 0; sec < fNumSec; sec++)
+    {
+        fTwimCal[sec]->Init();
+    }
+    for (Int_t i = 0; i < nHits; i++)
+    {
+        mappedData[i] = dynamic_cast<R3BTwimMappedData*>(fTwimMappedDataCA->At(i));
```
YanzhaoW: I'm confused about why we need a list of pointers to the mappedData. If I replace it with
```cpp
auto* mappedData = dynamic_cast<R3BTwimMappedData*>(fTwimMappedDataCA->At(i));
```
I don't see any problem with this.

twim/calibration/R3BTwimMapped2Cal.cxx
```diff
@@ -191,97 +236,86 @@ void R3BTwimMapped2Cal::Exec(Option_t*)
...
+        auto secId = mappedData[i]->GetSecID() - 1;
+        auto anodeId = mappedData[i]->GetAnodeID() - 1;
+        if (anodeId < fNumAnodes && fCal_Par->GetInUse(secId + 1, anodeId + 1) == 1)
         {
-            mulanode[s][i] = 0;
-            for (Int_t j = 0; j < fMaxMult; j++)
+            auto pedestal = CalEParams[secId]->GetAt(fNumEParams * anodeId);
+            auto slope = CalEParams[secId]->GetAt(fNumEParams * anodeId + 1);
+            fTwimCal[secId]->GetAnode(anodeId)->SetVal(pedestal + slope * mappedData[i]->GetEnergy(),
+                                                       mappedData[i]->GetTime());
```
YanzhaoW: If I understand correctly, the fTwimCal is first zero initialised with a maximal amount of elements. And then fill the valuable values according to section ids and anode ids. 

If so, I would suggest to use std::map in your data structure:
```cpp
class CalSection {
    public:
    //...
    using AnodeMap = std::map<size_t, CalAnode>;
    private:
        AnodeMap fAnode;
        AnodeMap fTref;
};

std::map<size_t, CalSection> fTwimCal;
```

Latter when you fill the value:
```cpp
        auto secId = mappedData[i]->GetSecID() - 1;
        auto anodeId = mappedData[i]->GetAnodeID() - 1;
        
        auto anodes = AnodeMap{};
        auto anode = CalAnode{};
        anode.SetE();
        //...
        anodes.insert(anodeID, anode);

        //...
        fTwimCal.insert(sectionId, anodes);
```

ryotani: Thanks for the suggestion. std::map sounds interesting solution, but I need to handle multiple hits for the respective channel, thus the energy and timing information should be stored in array or vector format. Do I understand correctly that the example you showed can overwrite the numbers if there is more than one hit in a channel?

twim/calibration/R3BTwimMapped2Cal.cxx
ryotani: I totally agree with this suggestion. Someone who wrote this class defined the array but I think just a pointer within the loop should be sufficient.

twim/calibration/R3BTwimMapped2Cal.cxx
YanzhaoW: Ok, in this case you need to have a check whether the AnodeMap has been inserted before. If no, insert it. If yes, just get it:

```cpp
if(fTwimCal.find(sectionID) != fTwimCal.end())
{
    fTwimCal.insert({sectionID, AnodeMap{}});
}
auto anodes = fTwimCal.at(sectionID);
if(anodes.find(anodeID) != anodes.end())
{
     anodes.insert({anodeID, CalAnode{}});
}
anodes.at(anodeID).GetHits().push_back(Hit);
```
This way is very memory efficient but not very good at performance (anyway, there is always a tradeoff between the performance and memory efficiency).

twim/calibration/R3BTwimMapped2Cal.h
```diff
@@ -88,13 +94,55 @@ class R3BTwimMapped2Cal : public FairTask
...
+            fE.at(fmult) = ene;
+            fDT.at(fmult) = dtime;
+            fmult++;
+        };
+        auto GetMult() const { return fmult; };
+        auto GetE(Int_t mult) { return fE.at(mult); };
+        auto GetDT(Int_t mult) { return fDT.at(mult); };
+
+      private:
+        Int_t fmult;
```
YanzhaoW: I remember that we had a consensus that c++ standard types (such as int, double, int8_t, etc) are preferred to ROOT types (Int_t, Double_t, etc). 

Please also don't forget to inline initialize those literal types.
```cpp
private:
    int fmult = 0;
```

twim/calibration/R3BTwimMapped2Cal.h
ryotani: Thanks, this change helped me to avoid using new/make_unique!

ryotani:
Sorry for making many attempts for the pull requests. I think it passed all the tests now. Should I rebase to the latest dev branch, or can it be merged with no problem?

twim/calibration/R3BTwimMapped2Cal.h
ryotani: Oh sorry, I missed this comment. Following the suggestion by clang-tidy, I put a line to initialise the variable with
`Int_t fmult { 0 };`
Do you think I should modify all the ROOT types to C++ standard types?

YanzhaoW: No, I don't think that's necessary. Just keep in mind that c++ standard types should be used when writing new code.

YanzhaoW:
Yes, of course. If there is no other issue and you are ready for the merging, please rebase to the latest dev branch:
```shell
# at your PR branch and your dev branch is the latest
git rebase dev
```
And don't forget to squash all commits into one:
```shell
git rebase -i HEAD~${Number of commits to be squashed}
```

ryotani:
Thanks for the help @YanzhaoW. I rebased to the latest commit in the dev branch. Then squash merged two commits.

ryotani:
Hi @jose-luis-rs

This PR is also ready to be merged. Thanks.

jose-luis-rs:
Thanks @ryotani

PR #871, opened 2023-07-08
Rewrite R3BFileSource

YanzhaoW:
* Rewrite whole R3BFileSource into new R3BFileSource2, which includes most essential functionalities and keep the same interfaces. Users just need to change the name from `R3BFileSource` to `R3BFileSource2`. RunID determination is not implemented as it should not be done in the file source, but rather in the file header or event header.
* Change namespace from `r3b` to `R3B`
* Add two exception classes. Their only difference from the C++ standard version is the exception message now includes the source location information (just like R3BLOG).

The new R3BFileSource2 has been tested with NeulandMap2CalPar and NeulandDigitizer.

Feel free to comment below if you have more suggestions. If there is no problem for the new file source, we could consider replacing the old file source with it in the future.

---

Checklist:

### DIFF
--- a/neuland/digitizing/R3BNeulandDigitizer.cxx
+++ b/neuland/digitizing/R3BNeulandDigitizer.cxx
@@ -80,12 +80,12 @@ InitStatus R3BNeulandDigitizer::Init()
 
     // Initialize control histograms
     auto const PaddleMulSize = 3000;
-    hMultOne = r3b::root_owned<TH1I>(
+    hMultOne = R3B::root_owned<TH1I>(
         "MultiplicityOne", "Paddle multiplicity: only one PMT per paddle", PaddleMulSize, 0, PaddleMulSize);
-    hMultTwo = r3b::root_owned<TH1I>(
+    hMultTwo = R3B::root_owned<TH1I>(
         "MultiplicityTwo", "Paddle multiplicity: both PMTs of a paddle", PaddleMulSize, 0, PaddleMulSize);
     auto const timeBinSize = 200;
-    hRLTimeToTrig = r3b::root_owned<TH1F>("hRLTimeToTrig", "R/Ltime-triggerTime", timeBinSize, -100., 100.);
+    hRLTimeToTrig = R3B::root_owned<TH1F>("hRLTimeToTrig", "R/Ltime-triggerTime", timeBinSize, -100., 100.);
 
     return kSUCCESS;
 }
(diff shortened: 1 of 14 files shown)


ryotani:
Hi, @YanzhaoW. Just out of curiosity, what is the motivation for the new class? Is it mostly to modernise the way of writing codes, or are there new functionalities included?
What shall we have to do if we use FileSource2? You changed the namespace to R3B, does it affect other codes also?

YanzhaoW:
Hi, @ryotani. Thanks for your comment.

Kind of both. The main motivation is to rewrite it in a clean way. The new version is way shorter and easier to understand.

You could see the difference by counting the lines of code:
`R3BFileSource.h`: 230
`R3BFileSource.cxx`: 1028
`R3BFileSource2.h`: 123
`R3BFileSource2.cxx`: 424

Actually in the beginning, what really set my teeth on edge is the strange interfaces of the old class. If I want to use multiple root files as the data source, I have to add the first file in the beginning and then add the rest of files together: 
```c++
    auto source = new R3BFileSource(TString::Format(inFilePattern, RunID));
    source->SetRunId(999);
    for (int i = 1; i <= nSubruns; i++)
    {
      source->AddFile(TString::Format(inFilePattern, RunID+i));
    }
```
This is strange. So I went to the source code to see whether there is an option to add them together. And what I found is a huge mess. :D So instead of wasting time on adding more mess on it. I thought it's better to rewrite it. So, with the new FileSource, you could just do:
```c++
auto source = new R3BFileSource2(); // DON't USE NEW
    for (int i = 0; i <= nSubruns; i++)
    {
      source->AddFile(TString::Format(inFilePattern, i)); // Use fmt::format()
    }
```
And the runID will be automatically detected and set from the root file. So no need to set the runID. There is also better error handling from the new class.

> What shall we have to do if we use FileSource2? You changed the namespace to R3B, does it affect other codes also?

The only thing you need to do is just to change the name `R3BFileSource` to `R3BFileSource2`. They should have the same interfaces. If not, please tell me and I will add it to the new class.

The namespace changing is only for the functions I made few months ago, such as `root_owned<>` or `make_rootfile<>`. I used the namespace `r3b`. But to make it consistent to other file names, `R3B` should be better. There is no namespace for the `R3BFileSource2`.

jose-luis-rs:
Nice work @YanzhaoW, if there are no objections, I will do the merge for testing it.

YanzhaoW:
Hi @jose-luis-rs Could you merge this PR?

PR #888, opened 2023-08-12
[HOT FIX] Bug fixing in Neuland digitisation simulation

YanzhaoW:
Thanks to @ivanalihtar and Nikhil, some bugs were found in the new NeuLAND digitizer classes.

Here are the changes to fix those bugs:

1. Make the global hit parameter non-const if user needs to read the parameter from a root file.
2. Add an option of runID in neulandSim executable such that the runID could have the same value as in the parameter root file.
3. Add additional information in the manual, explicitly saying the runIDs must be the same in both simulation file and parameter file.
4. Fix the mistake made in the distance calculation. Originally, the reconstructed signal distance is the negative of its original predefined value. For example, if a hit at position of 1 m is added to the paddle, the result of its corresponding signal has a position of -1 m.
5. Add an additional googletest for the distance checking, such that such mistake won't happen again. 

Feel free to comment below.

Checklist:

### DIFF
--- a/neuland/digitizing/R3BDigitizingPaddle.h
+++ b/neuland/digitizing/R3BDigitizingPaddle.h
@@ -15,7 +15,7 @@
 #define DIGITIZING_PADDLE_H
 
 #include "R3BDigitizingChannel.h"
-#include <FairLogger.h>
+#include <R3BLogger.h>
 #include <RtypesCore.h>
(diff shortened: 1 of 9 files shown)


YanzhaoW:
@jose-luis-rs Ready to be merged.

PR #920, opened 2024-01-11
Creating new data structures for ttt10

jose-luis-rs:
Update of TTT10 data structures and creating new ones for CAL and HIT data levels

---

Checklist:

### DIFF
--- a/r3bdata/ttt10Data/R3BTTTXCalData.cxx
+++ b/r3bdata/ttt10Data/R3BTTTXCalData.cxx
@@ -12,39 +12,32 @@
  ******************************************************************************/
 
 // -------------------------------------------------------------------------
-// -----                      R3BTTTXMappedData source file            -----
+// -----                  R3BTTTXCalData source file                   -----
 // -------------------------------------------------------------------------
 
-#include "R3BTTTXMappedData.h"
+#include "R3BTTTXCalData.h"
 #include <fmt/core.h>
 
-R3BTTTXMappedData::R3BTTTXMappedData(uint8_t DetID, uint8_t StripID, int32_t time, int32_t energy, bool pu, bool ov)
+R3BTTTXCalData::R3BTTTXCalData(uint8_t DetID, uint8_t StripID, double time, double energy)
     : fDetID(DetID)
     , fStripID(StripID)
     , fTime(time)
     , fEnergy(energy)
-    , fPileup(pu)
-    , fOverflow(ov)
 {
 }
 
-std::string R3BTTTXMappedData::toString() const
+std::string R3BTTTXCalData::toString() const
 {
-    return fmt::format("DetID: {}, StripID: {}, Time: {}, Energy: {}, Pileup: {}, Overflow: {}",
-                       GetDetID(),
-                       GetStripID(),
-                       GetTime(),
-                       GetEnergy(),
-                       GetPileupStatus(),
-                       GetOverflowStatus());
+    return fmt::format(
+        "DetID: {}, StripID: {}, Time: {}, Energy: {}", GetDetID(), GetStripID(), GetTime(), GetEnergy());
 }
 
-void R3BTTTXMappedData::Print(const Option_t*) const { std::cout << *this << std::endl; }
+void R3BTTTXCalData::Print(const Option_t*) const { std::cout << *this << std::endl; }
 
-std::ostream& operator<<(std::ostream& os, const R3BTTTXMappedData& data)
+std::ostream& operator<<(std::ostream& os, const R3BTTTXCalData& data)
 {
     os << data.toString();
     return os;
 }
 
-ClassImp(R3BTTTXMappedData)
+ClassImp(R3BTTTXCalData)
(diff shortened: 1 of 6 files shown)


r3bdata/ttt10Data/R3BTTTXCalData.h
```diff
@@ -0,0 +1,65 @@
...
+
+    // Support for printing
+    [[nodiscard]] std::string toString() const;
+    void Print(const Option_t*) const override;
+
+  protected:
+    uint8_t fDetID = 0;
+    uint8_t fStripID = 0;
+    int32_t fTime = 0;
+    int32_t fEnergy = 0;
```
ryotani: After the calibration, they should be double (or float). Can the default value being NAN?

r3bdata/ttt10Data/R3BTTTXCalData.cxx
```diff
@@ -0,0 +1,43 @@
...
+ ******************************************************************************/
+
+// -------------------------------------------------------------------------
+// -----                  R3BTTTXCalData source file                   -----
+// -------------------------------------------------------------------------
+
+#include "R3BTTTXCalData.h"
+#include <fmt/core.h>
+
+R3BTTTXCalData::R3BTTTXCalData(uint8_t DetID, uint8_t StripID, int32_t time, int32_t energy)
```
ryotani: int32_t --> double or float?

r3bdata/ttt10Data/R3BTTTXHitData.cxx
```diff
@@ -0,0 +1,50 @@
...
+ ******************************************************************************/
+
+// -------------------------------------------------------------------------
+// -----                  R3BTTTXHitData source file                   -----
+// -------------------------------------------------------------------------
+
+#include "R3BTTTXHitData.h"
+#include <fmt/core.h>
+
+R3BTTTXHitData::R3BTTTXHitData(uint8_t detID, double xpos, double energy, double angle, double charge, double time)
```
ryotani: Probably we don't need detID. I am thinking of using this Hit Level to store the angle deduced from two (or more) detectors. So the detID doesn't make sense at this stage.

r3bdata/ttt10Data/R3BTTTXHitData.h
```diff
@@ -0,0 +1,72 @@
...
+    inline void SetAngle(double angle) { fAng = angle; }
+    inline void SetChargeZ(double charge) { fCharge = charge; }
+    inline void SetTime(double time) { fTime = time; }
+
+    // Support for printing
+    [[nodiscard]] std::string toString() const;
+    void Print(const Option_t*) const override;
+
+  protected:
+    uint8_t fDetID = 0;
```
ryotani: fDetID is not needed at Hit Level

r3bdata/ttt10Data/R3BTTTXHitData.h
```diff
@@ -0,0 +1,72 @@
...
+    double fXpos = std::nan("");
+    double fEnergy = 0;
+    double fAng = std::nan("");
+    double fCharge = 0;
+    double fTime = 0;
```
ryotani: Why is the reason the default values not being nan for all? It will be anyway defined by cal2hit task though.

ryotani (commented):
Thanks @jose-luis-rs ! it looks great. I put some comments on the code. If you could take a look when you have a spare time, that would be great!

r3bdata/ttt10Data/R3BTTTXHitData.h
jose-luis-rs: Because charge = 0, energy=0 and time=0 mean that there is not a particle. Angle =0 or Xpos=0 could be even with ions passing through the detector

r3bdata/ttt10Data/R3BTTTXHitData.h
jose-luis-rs: Ok, I can remove it

r3bdata/ttt10Data/R3BTTTXHitData.cxx
jose-luis-rs: Ok, but the angle should be given only by the TTT10 detectors, the combination with other detectors for tracking is part of tracking tasks

r3bdata/ttt10Data/R3BTTTXCalData.cxx
jose-luis-rs: Uppss, I will change it

r3bdata/ttt10Data/R3BTTTXHitData.cxx
ryotani: Right. The angle here will be just ratios x-diff / z-diff between two detectors and will not give precise angles but still probably good to try to calculate...

r3bdata/ttt10Data/R3BTTTXHitData.h
jose-luis-rs: Everything is nan, this will remove all the zeros from your histograms later.

jose-luis-rs:
Ryo, let me know if you are happy with this version and then I will do the merge. Thanks

ryotani:
Thanks. Everything looks good.

PR #948, opened 2024-01-21
Implementation on NeuLAND T0 trigger (LMD reader and map2cal task)

YanzhaoW:
Use T0 trigger coarse time to check whether time jump occurs in NeuLAND.

### Method:

~~Check the T0 trigger coarse time value for every signal against the corresponding reference. Correction is made on coarse time value (both pmt signals and trigger signals) according to:~~
```c++
correction = static_cast<int>(std::remainder(t0_value - t0_ref, 1U << 11U )); // 1U << 11U is TAMEX full cycle
```
~~The reference value is the t0 value of the first event of the corresponding module number in the analysis run.
See R3BNeulandT0Checker for more detail.~~

### TODO

1. Waiting for implementation of NeuLAND in ucesb (upexps) side for the full testing.
2. Waiting for the specification of the map between T0 module numbers and signal module numbers.

Any suggestions and questions are welcomed. 

@igasparic @leandrofloyd @AndreaHorvat 

---

Checklist:

### DIFF
--- a/neuland/calibration/R3BNeulandMapped2Cal.cxx
+++ b/neuland/calibration/R3BNeulandMapped2Cal.cxx
@@ -103,6 +103,10 @@ InitStatus R3BNeulandMapped2Cal::Init()
     {
         LOG(info) << "Branch NeulandTrigMapped not found";
     }
+    if (trigger0_data_.try_init())
+    {
+        is_trig0_enabled = true;
+    }
 
     mgr->Register("NeulandCalData", "Neuland", fCal, kTRUE);
     fCal->Clear();
@@ -165,6 +169,14 @@ void R3BNeulandMapped2Cal::Exec(Option_t* option)
         if (header->GetTrigger() != fTrigger)
         {
             return;
+        }
+    }
+
+    if (is_trig0_enabled)
+    {
+        for (const auto& [module_num, trig0_data] : trigger0_data_)
+        {
+            t0_checker_.Add(module_num, trig0_data.GetCoarseTime1LE());
         }
     }
 
@@ -312,7 +324,9 @@ void R3BNeulandMapped2Cal::MakeCal()
         if (fWalkEnabled)
             timeLE = timeLE + WalkCorrection(qdc);
 
-        new ((*fCal)[fNPmt]) R3BNeulandCalData((iPlane - 1) * 50 + iBar, iSide, timeLE, trig_ns, qdc);
+        auto t0_corr = get_t0_correction((iPlane - 1) * 50 + iBar, iSide);
+        new ((*fCal)[fNPmt])
+            R3BNeulandCalData((iPlane - 1) * 50 + iBar, iSide, timeLE + t0_corr, trig_ns + t0_corr, qdc);
         fNPmt += 1;
 
         // Subtract trigger time.
@@ -327,6 +341,24 @@ void R3BNeulandMapped2Cal::MakeCal()
         // timeTE = fmod(timeTE + c_range + c_range/2, c_range) - c_range/2;
         // qdc = fmod(qdc + c_range + c_range/2, c_range) - c_range/2;
     }
+}
+
+auto to_t0_module_num(int signal_module_num, int side) -> int
+{
+    // TODO: placeholder
+    return 0;
+}
+
+auto R3BNeulandMapped2Cal::get_t0_correction(int module_num, int side) -> int
+{
+    auto t0_module_num = to_t0_module_num(module_num, side);
+    auto t0_entry = trigger0_data_.get().find(t0_module_num);
+    if (t0_entry == trigger0_data_.get().end())
+    {
+        LOG(error) << "cannot find trigger0 data corresponding to module number: " << module_num;
+        return 0;
+    }
+    return t0_checker_.GetCorrection(t0_module_num, t0_entry->second.GetCoarseTime1LE());
 }
 
 void R3BNeulandMapped2Cal::FinishEvent()
--- a/neuland/calibration/R3BNeulandMapped2Cal.h
+++ b/neuland/calibration/R3BNeulandMapped2Cal.h
@@ -15,7 +15,10 @@
 #define R3BNEULANDMAPPED2CAL_H
 
 #include "FairTask.h"
+#include "R3BPaddleTamexMappedData.h"
 #include "TH2F.h"
+#include <R3BIOConnector.h>
+#include <R3BNeulandT0Checker.h>
 
 class TClonesArray;
 class R3BTCalModulePar;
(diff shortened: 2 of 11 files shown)


YanzhaoW:
Hi, @michael-heil

As we have discussed in the analysis meeting today (sorry if I mistake you with another person), I have further questions on T0 value on signal coarse time correction.

### Question 1:
You mentioned we need to compare the T0 value between different modules. Let's suppose we calculate differences of T0 values of all NeuLAND cards to card 0. Thus, we would have difference values T0_00 (=0), T0_01, T0_02, T0_03 for events where T0 values are available. But to check the actual time jump happens, do we still need to check whether those T0 differences changed compared to the last available event?

### Question 2:
With T0 values available, I'm still not sure how exactly we implement these on the signal coarse time values. Here is my current thought:
with T0 values we could subtract it in the map level before the time calibration:
```text
neuland_signal_coarse_value_t0corrected = neuland_signal_coarse_value - neuland_t0_coarse_value
neuland_trigger_coarse_value_t0corrected = neuland_trigger_coarse_value - neuland_t0_coarse_value
```
for each event. And this could correctly eliminate time jump problem on coarse time.
Is this a correct way?

### Question 3:
Without T0 values, Neuland hit level time value would be:
```text
hit_time = (neuland_signal_time - neuland_trigger_time) - (los_signal_time - los_trigger_time)
```
and in the end, we calculate the global time offset (neuland_trigger_time - los_trigger_time) by using the gamma ray peak on the histogram over large amount of events. But this is based on the assumption that the global time offset is stable for each event. However, now we have this T0 value to consider, does that mean we also need to correct cal2hit process as well?
```text
hit_time = (neuland_signal_time - neuland_trigger_time) - (los_signal_time - los_trigger_time) - (neuland_t0_value - los_t0_value)
```
 
----------------------------------------------------
I would be really appreciated if other people from DAQ team could help this. @inkdot7 @klenze

inkdot7:
Dear @YanzhaoW ,

with `xx` as any detector with a 5 ns coarse 2^n counter, (e.g. `neuland` or `los` ...), an absolute coarse hit time (modulo the counter range)  should be:
```
xx_signal_coarse_value_t0corrected = xx_signal_coarse_value - xx_t0_coarse_value
```
Since the time measured now is absolute, the trigger time is no longer needed.  (It can be treated analogously however, and used as a check, since it is the same signal measured in all systems, so should for any pair of signals always show the same difference in any event.)

To get the precision time, the fine time of the channel would be added (or subtracted, depending on how it is measured):
```
xx_signal_value_t0corrected = xx_signal_coarse_value_t0corrected +/- xx_signal_fine_value
```

The values `xx_signal_value_t0corrected` for whatever channel pair in the setup (that use this) can then be directly subtracted to obtain their time difference.  (Plus the respective calibration offsets to handle cable lengths etc.)

Note: some events will have (all) their times close to the wrap of the coarse counters (i.e. some small, and some close to 2^n, n being the number of coarse bits).  To avoid having then time differences that need modulo-subtraction of 5 ns * 2^n, an additional event-wise coarse offset could be applied to have them end up around mid-coarse range instead.  With the same coarse offset applied to all channels, no difference is affected.  (It could also be chosen to be either 0 or 2^(n-1), i.e. just two values.)

Also note the *necessity* to consider the 'rough' fine time of the T0 signal when determining the `xx_t0_coarse_value`.  Otherwise it may itself be jumping for channels where the T0 signal (by bad luck) is close to the 5 ns wrap of the fine time range.  Such channels would then *always* be prone to jump.  ('rough' fine time is to determine the location to at least 'quadrant-precision' of the T0 'phase' inside the fine time range.)

Are there any verifications that the `xx_t0_coarse_value` actually works as advertised, i.e. that course counter differences can be stable even with partial coarse counter resets (e.g. power cycling of one system but not the other)?  (Based on directly unpacked data from the DAQ.)  I did not find anything in e.g. the elog*s* (two/three of them even used at the same time...), but perhaps did not look carefully enough.

igasparic:
I am not sure if I understand but what is the advantage of T0 compared to the trigger time? I would say the trigger time (as we have now) is better because it is distributed to all TAMEX cards whereas T0 goes to only one card per DAQ PC. So if a small 5ns jump appears in a card without T0 we cannot automatically correct it but we have to recalibrate the time offsets.

YanzhaoW:
@inkdot7 

Thanks for your detailed explanation.

In the email, you mentioned:

> The Exploder WRTCLK firmware, which has access to the WR time scale, can now generate a signal at times corresponding to absolute coarse counter values that are multiples of 2^n.  I.e., the occurrence of the pulse is calculated from the global time scale to happen at fixed 5*2^n ns times, not from a distributed (reset-like) signal.  It is thus repeatable after restarts, thunderstorms, and other reconnections. For now, n is chosen as 14, so the interval is 81920 ns.

T0 signal is sent to each detector at a period of 2^14 clock cycles (200MHz). However, the period of NeuLAND TAMEX period is 2^11 clock cycles. This means **not every period of NeuLAND clock could have a T0 value available**. And (if I'm not wrong) this could lead to **not every NeuLAND event would have a T0 value available**. In such cases, we still need a T0 value to calculate the `coarse_value_t0corrected` for those events without T0 values. 

So how do we infer a correct T0 value for those events where there is no T0 value?

inkdot7:
> So how do we infer a correct T0 value for those events where there is no T0 value?

@YanzhaoW 

The T0 coarse counter value, for those few events that it is measured, will always be the same.

Example with a bitmask:  With 2^11, the wrap is at `0x800` and such a coarse counter can take any value in the range `[0,0x7ff]`.  The T0 wrap at 2^14 is `0x4000`.  The T0 signal is emitted every time the global time reaches `0x4000 * 5` ns, without any offset,repeatably - since it is derived from a/the global (WR) time counter.

The coarse counter of each module will have an arbitrary offset (non-changing - unless there are jumps or resets).  In this case in the range `[0,0x7ff]`.   With that offset known, a coarse counter could be said after correction to have been reset at a global multiple of `0x0800 * 5` ns, i.e. give an absolute time (modulo `0x0800 * 5` ns).  Measuring the T0 signal will give that offset, every time it is measured.  So it needs to be remembered (per module).

Note the need to measure T0 (once in a while) in each=every module, to addresses the question of @igasparic above.  Indeed, if only measured in some modules, any other are not 'protected' against having to recalibrate after time jumps.

YanzhaoW:
@inkdot7 

Sorry this late reply. I've been following the conversation on our mailing and trying to understand as much as I can (though still not much :D). But I did have one concern in terms of the analysis workflow:

> Measuring the T0 signal will give that offset, every time it is measured. So it needs to be remembered (per module).

Events, representing separate physics interactions, can be independent with each other. It does't matter (as far as I know for NeuLAND) whether we analyze the 100th event first or 200th event first. This opens up a very important possibility that those events can be analysed **asynchronously** and thus the analysis process could speed up with a large factor in our HPC infrastructure.  Now if we need to remember T0 values for each module, we have to analyse events in a correct order. Such restriction may seem very little especially due to the fact that we have been always analysing events synchronously (either in a single core or with multi processes across multiple cores). But once we begin to push the algorithm to be asynchronous for better performance in couple of years, this restriction could become a huge downgrade.

Therefore in my opinion, it would be better in a long term if coarse time cycles are already corrected (in lmd data level) before the analysis process.

inkdot7:
@YanzhaoW indeed, keeping track of things between events is not pleasant.  Even with a serial analysis, it would not be possible to handle the first events in a file until the T0 appears.  Would not be good.

For TAMEX, with firmware support, the correction would already be applied as the coarse counter has adjusted in the FPGA.

For the VFTX, the DAQ software would know the correction value (as it certainly shall keep track and alert the user when a jump is detected) and thus can write that as an additional word (per module) in every event.  (Data-size wise this is not very much compared to other overgrown formats.)  Implementing this warning already now would be good.

klenze:
Counterpoint: from my understanding, we already have to go over the TDC data twice to get the fine time calibration. We can just discover the 5ns jumps then. 

(Of course, it would be more elegant if we could analyze data in a single pass, perhaps even discarding data until we have a calibration. For the envisioned T0 scheme, this would take up to 100ms, which I consider acceptable when analyzing minute-long LMD files. (Hot take: anything which avoids users having to deal with FairParams is worth 1% of statistics.) I guess obtaining a fine time calibration would take a lot longer?)

YanzhaoW:
> Counterpoint: from my understanding, we already have to go over the TDC data twice to get the fine time calibration. We can just discover the 5ns jumps then.
> 
> (Of course, it would be more elegant if we could analyze data in a single pass, perhaps even discarding data until we have a calibration. For the envisioned T0 scheme, this would take up to 100ms, which I consider acceptable when analyzing minute-long LMD files. (Hot take: anything which avoids users having to deal with FairParams is worth 1% of statistics.) I guess obtaining a fine time calibration would take a lot longer?)

Hi,

well, not entirely. For NeuLAND calibration TDC values will be calculated from the cosmic data (off-spill data once). And then the calibration parameters will be applied to the on-spill data. So we do need to go twice over TDC calculation, but on different data.

YanzhaoW:
I'm closing this PR as a better way seems to implement this correction before the analysis process.

PR #952, opened 2024-01-23
R3BTttxMapped2Cal is added

ryotani:
TTTXReader is renamed to TttxReader
Mapped2Cal is updated to reconstruct the calibrated energy and the relative timing to trigger. The initial values for Mwpc hit postion is set to NAN.

Describe your proposal.

---

Checklist:

* [X] Followed the [pull request guidlines](https://opensource.creativecommons.org/contributing-code/pr-guidelines/) and the [Git workflow](https://github.com/AnarManafov/GitWorkflow/blob/master/GitWorkflow.markdown)


mwpc/mwpc3/R3BMwpc3Cal2Hit.cxx
```diff
@@ -157,7 +157,7 @@ void R3BMwpc3Cal2Hit::S467()
         Int_t padId = 0;
         Int_t padmx = -1, padmy = -1;
         Double_t q = 0., qmx = 0., qmy = 0., qleft = 0., qright = 0., qdown = 0., qup = 0.;
-        Double_t x = -1000., y = -1000.;
```
jose-luis-rs, line 160: This change is not needed because you don't need the mwpc3, right?

ryotani:
@YanzhaoW @jose-luis-rs Hi, I need your support. I wanted to modify a single line in mwpc/mwpc3/R3BMwpc3Cal2Hit.cxx then I got an error pointed by clang-tidy. This is because the Mwpc3 class was written by someone previously to include SofTofw classes, which are in another repository. Could you tell me how I can modify this code not to have such issues?

mwpc/mwpc3/R3BMwpc3Cal2Hit.cxx
ryotani: Yes, for this experiment but I would like to apply this anyway for the offline analysis in the previous experiments. Is it difficult to solve the clang-tidy issues?

audreychatillon:
there will be 3 Mw in the setup.
One at the entrance of cave C and two just before GLAD.

The two first are the ones segmented Up/Down, the third one is the high
precision - Y Mw

mwpc/mwpc3/R3BMwpc3Cal2Hit.cxx
jose-luis-rs: I can accept it directly because during the compilation of r3broot this class is only used if the sofia repo is added.

jose-luis-rs:
Yes, but MWPC3 corresponds to the big one, the small mwpcs are 0, 1 and 2.
If the rest of code is OK I will merge it directly. Don't worry!!

YanzhaoW:
@ryotani Could you just delete `#include "R3BSofTofWHitData.h"` in R3BMwpc3Cal2Hit.h? It seems that this header file is not needed at all.

ryotani:
@YanzhaoW If I compile without including R3BSofTofWHitData, then it fails. The Data class is used in the function `ReconstructHitWithTofWallMatching()`.

YanzhaoW:
Very strange. It seems like R3BMwpc3Cal2Hit.cxx isn't even compiled. Therefore the whole compilation of R3BRoot is still successful even though that header file is missing. it also passes the CI because there is no test using this file neither. Have you tried to use a program that relies on R3BMwpc3Cal2Hit.cxx without Sofia repository?

This is a serious bug. R3BRoot, by no means, need to be relied on a sub repositories like Sofia.  You either copy all necessary files from Sofia to R3BRoot or put R3BMwpc3Cal2Hit to Sofia repository.

The file was written by @jose-luis-rs in [this PR](https://github.com/R3BRootGroup/R3BRoot/pull/504). Maybe he should comment on this.

YanzhaoW:
I see why it passed CI:

https://github.com/R3BRootGroup/R3BRoot/blob/aa463a3702b9838885287fa5c9218f66df53b2cf/mwpc/CMakeLists.txt#L47-L55

Everything in mwpc3 folder is not compiled. You may just delete everything over there.

jose-luis-rs:
> The file was written by @jose-luis-rs in [this PR](https://github.com/R3BRootGroup/R3BRoot/pull/504). Maybe he should comment on this.

This is because mwpc3 is only used for the analysis with the sofia repository and thus this class is only compiled if the sofia repo exists. Initially the mwpcs were implemented only for the sofia experiments, but then our collaborators saw that these detectors could be a good option for the tracking of R3B, so we moved the full forder sofia/mwpc into R3BRoot, including mwpc3, because it was much easier.

ryotani:
I compiled the PR'ed version without sofia sub-repositoy. It compiled with no problem. As you pointed out, I think the mwpc/CMakeLists.txt decides whether it includes mw3. mw3 was (and will be?) only used with SofiaTofW, so this should be no problem.

https://github.com/R3BRootGroup/R3BRoot/blob/aa463a3702b9838885287fa5c9218f66df53b2cf/mwpc/CMakeLists.txt#L67-L69

audreychatillon:
Mwpc3 is only size compatible with the SofTofw, which is never used in the
standard R3B experiment.

These 2 detectors have only been used for the fission experiments.

Le mar. 23 janv. 2024, 17:48, Ryo Taniuchi ***@***.***> a
écrit :

PR #1063, opened 2024-03-21
Updated IncomingBeta reconstruction for s091 experiment.

ryotani:
Clean-up and minor improvements in los classes.

Include the timing of the S2 hit after subtraction of the reference time in FrsSciPosCalData.

Bug fix in R3BFrsSciTcal2Cal when obtaining good pairs of hits of left and right PMTs.

Update IncomingData to reconstruct S2(FrsSci) to LosCal for the s091 experiment. Los CFD out was used for the Tref signal for S2, so LosCal is used for reconstructing ToF specific to the experiment. When multiple hits in Los are seen, only good hits wrt the trigger timing are chosen.

Describe your proposal.

---

Checklist:

* [X] Rebased against `dev` branch
* [X] My name is in the resp. CONTRIBUTORS/AUTHORS file
* [X] Followed the [pull request guidlines](https://opensource.creativecommons.org/contributing-code/pr-guidelines/) and the [Git workflow](https://github.com/AnarManafov/GitWorkflow/blob/master/GitWorkflow.markdown)
* [X] Followed the [seven rules of great commit messages](https://chris.beams.io/posts/git-commit/#seven-rules)

### DIFF
--- a/analysis/CMakeLists.txt
+++ b/analysis/CMakeLists.txt
@@ -66,6 +66,7 @@ ${R3BROOT_SOURCE_DIR}/r3bdata/tofiData
 ${R3BROOT_SOURCE_DIR}/r3bdata/pdcData
 ${R3BROOT_SOURCE_DIR}/r3bdata/pspData
 ${R3BROOT_SOURCE_DIR}/r3bdata/fibData
+${R3BROOT_SOURCE_DIR}/r3bdata/frssciData
 ${R3BROOT_SOURCE_DIR}/r3bdata/ptofData
 ${R3BROOT_SOURCE_DIR}/r3bdata/sampData
 ${R3BROOT_SOURCE_DIR}/r3bdata/sfibData
--- a/analysis/offline/R3BAnalysisIncomingID.h
+++ b/analysis/offline/R3BAnalysisIncomingID.h
@@ -101,6 +101,8 @@ class R3BAnalysisIncomingID : public FairTask
     // Acsessor to set use of trigger corrected times
     void SetUseTref() { fUseTref = kTRUE; }
 
+    void SetNumDet(int val) { fNumDet = val; }
+
   protected:
     R3BEventHeader* fHeader{}; // Event header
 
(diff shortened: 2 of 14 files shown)


nmozumdar:
Hi Ryo, just looking through it fast, your updates look ok from my side. However I should give you a heads up that in the next step you will have to match the Musli hit for the charge to the corresponding los hit. Which means if you take your current approach to selecting the hit within the task you will have to copy the same stuff multiple times. In fact it will be even more complicated within the Musli Task. This is why I took the approach of selecting the hit within the TProvideTStart and then matching it to the musli hit hence selecting for the correct Frs Z. Please have a look at he MusliMapped2Cal task.

ryotani:
Thanks @nmozumdar for the comment. Actually, your point is one of the difficult parts for me to write down in the code.

As you may remember, the trigger timing distributed to the DAQ systems was one of the big confusions during the experiment last month. The tref signal sent to S2 was for some reason having a jittering. Thus we should live with another trigger reference signal; that is a CFD out from a channel of the LOS PMT. While TStart is deduced by the difference from the average timing of 8 PMTs in LOS to the trigger timing, the definition of reference timing for S2 DAQ is slightly different. As I didn't want to change TStartProvider too much, I decided to code in the R3BIncomingBeta instead.
In the actual analysis for this experiment, I think one can set the same condition in TStartProvider with the SetWindow function to select the same LOS hit for all the analysis:
https://github.com/R3BRootGroup/R3BRoot/blob/4f4b89370d2bbf0c0deb25aec87f6394542381be/los/R3BLosProvideTStart.h#L34-L39

ryotani:
@jose-luis-rs Hi, this is also fine to be merged.

jose-luis-rs:
Hello @ryotani , did you check that the code still works fine for the S509 PID?

jose-luis-rs:
for example for run0190

ryotani:
@jose-luis-rs I fixed a bug. Now it should be alight with s509.

jose-luis-rs:
Thanks @ryotani 
Please, could you squash your commits?

ryotani:
@jose-luis-rs Is this fine?

jose-luis-rs:
> @jose-luis-rs Is this fine?

Yes, thanks

PR #1276, opened 2025-09-26
Updates for CALIFA ToT calibration

Mrunmoy-Jena:
### Updates for CALIFA ToT calibration

-**R3BCalifaTotCalPar.h**: Minor logic flaw fixed in SetCryCalParams() function, now sets ToT parameter at correct index
-**R3BCalifaMapped2CrystalCal.cxx** : Minor update to include some default ToT calibration parameters (when no ToT parameters found in corresponding .par file)
-**R3BCalifaCrystalCalData.h** : Added a GetCalEnergy() that returns only the calibrated energy (and not the extrapolated energy), useful for making correlation plots of CalEnergy v/s ToTEnergy for ToT QC plots.
-**R3BCalifaCrystalCal2TotCalPar.h**: Major updates; introduced new member functions and variables to achieve better control over ToT v/s energy fits and store them
-**R3BCalifaCrystalCal2TotCalPar.cxx**: Major updates; logic to isolate 'main band' and 'shadow band' in these plots, and fit them separately

### DIFF
--- a/califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
+++ b/califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
@@ -16,22 +16,23 @@
 #include <FairRunAna.h>
 #include <FairRuntimeDb.h>
 
-#include "R3BCalifaCrystalCal2TotCalPar.h"
-#include "R3BCalifaCrystalCalData.h"
-#include "R3BCalifaMappingPar.h"
-#include "R3BCalifaTotCalPar.h"
-
 #include <TClonesArray.h>
-#include <TF1.h>
-#include <TGraph.h>
-#include <TH1F.h>
+#include <TFile.h>
+#include <TLegend.h>
 #include <TMath.h>
 #include <TObjArray.h>
+#include <TROOT.h>
 #include <TRandom.h>
 #include <TSpectrum.h>
 #include <TVector3.h>
 #include <iostream>
 #include <stdlib.h>
+
+#include "R3BCalifaCrystalCal2TotCalPar.h"
+#include "R3BCalifaCrystalCalData.h"
+#include "R3BCalifaMappedData.h"
+#include "R3BCalifaMappingPar.h"
+#include "R3BCalifaTotCalPar.h"
 
 R3BCalifaCrystalCal2TotCalPar::R3BCalifaCrystalCal2TotCalPar()
     : R3BCalifaCrystalCal2TotCalPar("R3BCalifaCrystalCal2TotCalPar", 1)
@@ -40,29 +41,80 @@ R3BCalifaCrystalCal2TotCalPar::R3BCalifaCrystalCal2TotCalPar()
 
 R3BCalifaCrystalCal2TotCalPar::R3BCalifaCrystalCal2TotCalPar(const char* name, Int_t iVerbose)
     : FairTask(name, iVerbose)
-    , fMap_Par(NULL)
-    , fTotCal_Par(NULL)
-    , fCrystalCalDataCA(NULL)
-    , fNumCrystals(1)
+    , fDebugMode(true)
+    , fNumCrystals(5088)
     , fNumParam(2)
-    , fMinStadistics(100)
-    , fThreshold(0)
-    , fLeft(500.)
-    , fRight(1400.)
-    , fDebugMode(false)
+    , fMinStatistics(20)
+    , fThreshold(100)
+    , fLeft(10000.)
+    , fRight(25000.)
+    , fLeftProton(130000.)
+    , fRightProton(250000.)
+    , fa0(6000.)
+    , fa1(1000.)
+    , fa0proton(60000.)
+    , fa1proton(1000.)
+    , fMap_Par(nullptr)
+    , fTotCal_Par(nullptr)
+    , fCrystalCalDataCA(nullptr)
+    , fMappedDataCA(nullptr)
+    , tot_vs_energy_hist()
+    , totCanvases()
+    , fGraphs1()
+    , fGraphs2()
+    , fFits1()
+    , fFits2()
 {
 }
 
 R3BCalifaCrystalCal2TotCalPar::~R3BCalifaCrystalCal2TotCalPar()
 {
-    if (fCrystalCalDataCA)
-        delete fCrystalCalDataCA;
+    // Clean up TGraphErrors and TF1 objects first
+    for (Int_t i = 0; i < fGraphs1.size(); i++)
+    {
+        if (fGraphs1[i])
+        {
+            delete fGraphs1[i];
+            fGraphs1[i] = nullptr;
+        }
+        if (fGraphs2[i])
+        {
+            delete fGraphs2[i];
+            fGraphs2[i] = nullptr;
+        }
+
+        delete fFits1[i];
+        delete fFits2[i];
+        fFits1[i] = nullptr;
+        fFits2[i] = nullptr;
+    }
+
+    // Clean up ToT_vs_energy hists
+    for (Int_t j = 0; j < tot_vs_energy_hist.size(); j++)
+    {
+        if (tot_vs_energy_hist[j])
+        {
+            delete tot_vs_energy_hist[j];
+            tot_vs_energy_hist[j] = nullptr;
+        }
+    }
+
+    for (Int_t k = 0; k < totCanvases.size(); k++)
+    {
+        if (totCanvases[k])
+        {
+            delete totCanvases[k];
+            totCanvases[k] = nullptr;
+        }
+    }
+
+    fCrystalCalDataCA = nullptr;
+    fMappedDataCA = nullptr;
 }
 
 void R3BCalifaCrystalCal2TotCalPar::SetParContainers()
 {
     // Parameter Container
-    // Reading califaMappingPar from FairRuntimeDb
     FairRuntimeDb* rtdb = FairRuntimeDb::instance();
     if (!rtdb)
     {
@@ -84,16 +136,17 @@ void R3BCalifaCrystalCal2TotCalPar::SetParameter()
     if (!fMap_Par)
     {
         LOG(error) << "R3BCalifaMapped2CrystalCalPar::Container califaMappingPar not found.";
+        return;
     }
     //--- Parameter Container ---
     fNumCrystals = fMap_Par->GetNumCrystals(); // Number of crystals x 2
     LOG(info) << "R3BCalifaMapped2CrystalCalPar::NumCry " << fNumCrystals;
-    // fMap_Par->printParams();
 }
 
 InitStatus R3BCalifaCrystalCal2TotCalPar::Init()
 {
-    LOG(info) << "R3BCalifaCrystalCal2TotCalPar::Init()";
+    LOG(info) << "R3BCalifaCrystalCal2TotCalPar::Init() called";
+    gROOT->SetBatch(kTRUE);
 
     FairRootManager* rootManager = FairRootManager::Instance();
     if (!rootManager)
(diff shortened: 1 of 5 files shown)


califa/pars/R3BCalifaTotCalPar.h
```diff
@@ -56,8 +56,8 @@ class R3BCalifaTotCalPar : public FairParGenericSet
     inline void SetNumParametersFit(int numberParams) { fNumParamsFit = numberParams; }
     inline void SetCryCalParams(Float_t cc, int cry)
     {
-        assert(std::clamp(cry, 1, fNumCrystals) == cry && "Number of crystal out of range");
-        fCryCalParams->AddAt(cc, cry);
+        assert(std::clamp(cry, 1, fNumCrystals * fNumParamsFit) == cry && "Number of crystal out of range");
```
hapol: Suggestion: Change the message in the assert to specify better the problem (as "Number of crystal Tot parameters out of range")

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
```diff
@@ -17,8 +17,27 @@
 #define R3BCalifaCrystalCal2TotCalPar_H 1
 
 #include <FairTask.h>
```
hapol, line 19: Please, check that included headers are really required (some like TF2 does not seem to be used anywhere). Remove those unused, please.
It is usually recommended to add the #include in the implementation file, as headers could be read by many  parts, while the includes are required for the object compilation.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
```diff
@@ -65,36 +77,59 @@ class R3BCalifaCrystalCal2TotCalPar : public FairTask
...
+
+    TH2D** tot_vs_energy_hist;
+    TCanvas** totCanvases;
+    std::vector<TGraphErrors*> fGraphs1;
+    std::vector<TGraphErrors*> fGraphs2;
+    std::vector<TF1*> fFits1;
+    std::vector<TF1*> fFits2;
 
   public:
     ClassDef(R3BCalifaCrystalCal2TotCalPar, 1);
```
hapol: The modification in the data structure may require a new version number in the ClassDef.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
```diff
@@ -65,36 +77,59 @@ class R3BCalifaCrystalCal2TotCalPar : public FairTask
...
+    Double_t fRightProton; // upper limit of calibrated energy [keV] to be used for ToT fits (proton range)
+    Double_t fa0, fa1, fa0proton, fa1proton; // Initial parameters for the ToT fit functions
+
+    R3BCalifaMappingPar* fMap_Par;   // Parameter container with mapping
+    R3BCalifaTotCalPar* fTotCal_Par; // Container for Tot-Cal parameters
+    TClonesArray* fCrystalCalDataCA; // Array with calibrated energy
+    TClonesArray* fMappedDataCA;     // Array with uncalibrated ToT
+
+    TH2D** tot_vs_energy_hist;
+    TCanvas** totCanvases;
```
hapol: If possible, get rid of Canvases at this stage, and move the Histograms and other ROOT graphical objects to a ROOT file, improving performance. The file can be open and analysed at the end.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
```diff
@@ -149,96 +241,487 @@ InitStatus R3BCalifaCrystalCal2TotCalPar::ReInit()
     return kSUCCESS;
 }
 
-void R3BCalifaCrystalCal2TotCalPar::Reset() {}
+void R3BCalifaCrystalCal2TotCalPar::Reset()
+{
+    //  Clear the CA structure
+    LOG(info) << "Clearing CalifaCrystalCalData Structure";
+    if (fCrystalCalDataCA)
+        fCrystalCalDataCA->Clear();
```
hapol: Why do you need to reset the TClonesArray fCrystalCalDataCA? Where are you calling the Reset() function? I guess this should not be here, as you could Clear the data in a CA which is not under your ownership.

hapol (changes requested):
There a few comments along the code. The main requested change corresponds to the Reset() function and the Clear on a CA without ownership.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
```diff
@@ -65,36 +77,59 @@ class R3BCalifaCrystalCal2TotCalPar : public FairTask
...
+    Double_t fLeftProton;  // lower limit of calibrated energy [keV] to be used for ToT fits (proton range)
+    Double_t fRightProton; // upper limit of calibrated energy [keV] to be used for ToT fits (proton range)
+    Double_t fa0, fa1, fa0proton, fa1proton; // Initial parameters for the ToT fit functions
+
+    R3BCalifaMappingPar* fMap_Par;   // Parameter container with mapping
+    R3BCalifaTotCalPar* fTotCal_Par; // Container for Tot-Cal parameters
+    TClonesArray* fCrystalCalDataCA; // Array with calibrated energy
+    TClonesArray* fMappedDataCA;     // Array with uncalibrated ToT
+
+    TH2D** tot_vs_energy_hist;
```
pgrusell: As you do for fGraphs and fFits, you could also use std::vectors instead of raw double pointers (which is, in general, recommended). Is there any reason not to do this?

califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
```diff
@@ -18,51 +18,100 @@
...
+        {
+            delete fGraphs1[i];
+            fGraphs1[i] = nullptr;
+        }
+        if (fGraphs2[i])
+        {
+            delete fGraphs2[i];
+            fGraphs2[i] = nullptr;
+        }
+        fFits1[i] = nullptr;
```
pgrusell, line 88: Here you should delete the fFits1 and fFits2 as they are created by you with new.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
```diff
@@ -18,51 +18,100 @@
...
+    , fDebugMode(true)
+    , fNumCrystals(5088)
     , fNumParam(2)
-    , fMinStadistics(100)
-    , fThreshold(0)
-    , fLeft(500.)
-    , fRight(1400.)
-    , fDebugMode(false)
+    , fMinStatistics(20)
+    , idx(nullptr)
```
pgrusell: This variable (idx) is not used anywhere right?

califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
```diff
@@ -149,96 +241,487 @@ InitStatus R3BCalifaCrystalCal2TotCalPar::ReInit()
...
+    // Reset entries in output arrays, local arrays
+    // Reset();
+    auto nHits = fCrystalCalDataCA->GetEntriesFast();
+
+    if (nHits == 0)
         return;
 
     R3BCalifaCrystalCalData** CalHit = new R3BCalifaCrystalCalData*[nHits];
-    Int_t crystalId = 0;
+    R3BCalifaMappedData** MapHit = new R3BCalifaMappedData*[nHits];
```
pgrusell: This structure seems a little bit inefficient, why are you saving the hits in CalHit/MapHit if you are going to use them anywhere? In my opinion you could create a variable inside the loop like this:

```c++
auto* cal = dynamic_cast<R3BCalifaCrystalCalData*>(fCrystalCalDataCA->At(i));
auto* map = dynamic_cast<R3BCalifaCrystalMappedData*>(fCrystalMappedDataCA->At(i));
```

r3bdata/califaData/R3BCalifaCrystalCalData.h
```diff
@@ -68,6 +67,10 @@ class R3BCalifaCrystalCalData : public TObject
     [[nodiscard]] inline const double& GetNs() const { return fNs; }
     [[nodiscard]] inline const ULong64_t& GetTime() const { return fTime; }
     [[nodiscard]] inline const double& GetToTEnergy() const { return fToT_Energy; }
+    [[nodiscard]] inline const double& GetCalEnergy() const
```
pgrusell: Why do you return fEnergy (a double) by reference instead of by value?

califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
Mrunmoy-Jena: Yes, indeed its not used anywhere. Will remove it, thanks.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
Mrunmoy-Jena: Yes, I agree, was perhaps an oversight on my part. Will fix it.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
Mrunmoy-Jena: The logic is indeed that the fits and histograms are to be written to a ROOT file when this Tot2CalPar task is called. Would you suggest I move the declarations of these to the corresponding .cxx file?

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
Mrunmoy-Jena: So changing it to : ClassDef(R3BCalifaCrystalCal2TotCalPar, 2); should be okay ? Will do so.

califa/pars/R3BCalifaTotCalPar.h
Mrunmoy-Jena: Thanks, will implement this.

r3bdata/califaData/R3BCalifaCrystalCalData.h
Mrunmoy-Jena: I was simply keeping the definition similar to the already defined GetToTEnergy(), but yes I could change GetCalEnergy() to return by value instead.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
Mrunmoy-Jena: Sure, will do so.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.cxx
Mrunmoy-Jena: Yes, that makes sense. Modified it accordingly.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
Mrunmoy-Jena: Thanks, I will add these headers to the corresponding .cxx file and remove unnecessary ones.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
Mrunmoy-Jena: Yes I can do so. Again was following the old method where energy_vs_tot_crystal was implemented as double pointer. Have changed it now to vectors of type TH2D.

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
hapol: Yes, it will do. Thanks!

califa/calibration/R3BCalifaCrystalCal2TotCalPar.h
hapol: I was thinking in writing all graphical objects on a ROOT file with no TCanvas whatsoever, and therefore no graph output at the end of the class work. All histograms and graph are stored in the output file, in this way. But, as this is a class to obtain parameters (R3BCalifa*Par.h), it is not really important, as the class will not be included in the data flow for analysis.

pgrusell (approved):
Hi, 

Thanks for applying all the changes :) I don't know if @hapol has any further comment, but, from my side, the PR is ready for getting merged into dev.

Mrunmoy-Jena:
Hi,
Thanks a lot Pablo !

Regards
Mrunmoy
________________________________

hapol (approved):
Thank you!

PR #1337, opened 2026-04-20
feat(r3bgen): Added INCL-like vertex smearing and general Lorentz boost in CALIFATestGenerator

Mrunmoy-Jena:
r3bgen : Modified the R3BCALIFATestGenerator class to make simulation of gamma emission more realistic
- Add 3D vertex distribution (Gaussian in x,y and uniform in z) 
- Add a method to sample outgoing beam direction from a 2D Gaussian distribution (corresponding to angular divergence of outgoing fragment), so that photon theta is taken wrt to this outgoing fragment direction instead of the z axis; needed for accurate Doppler correction
- A general Lorentz boost is now applied in the direction of the outgoing fragment unit vector

### DIFF
--- a/r3bgen/R3BCALIFATestGenerator.cxx
+++ b/r3bgen/R3BCALIFATestGenerator.cxx
@@ -20,6 +20,7 @@
 #include <TMath.h>
 #include <TParticlePDG.h>
 #include <TRandom.h>
+
 #include <iomanip>
 #include <iostream>
 #include <sstream>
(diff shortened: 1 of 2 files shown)


r3bgen/R3BCALIFATestGenerator.cxx
```diff
@@ -1,15 +1,15 @@
...
- *                                                                            *
- *             This software is distributed under the terms of the            *
- *                 GNU General Public Licence (GPL) version 3,                *
- *                    copied verbatim in the file "LICENSE".                  *
- *                                                                            *
- * In applying this license GSI does not waive the privileges and immunities  *
- * granted to it by virtue of its status as an Intergovernmental Organization *
- * or submit itself to any jurisdiction.                                      *
- ******************************************************************************/
+/****************************************************************************
```
jose-luis-rs: Why did you change the header?

r3bgen/R3BCALIFATestGenerator.h
```diff
@@ -1,23 +1,29 @@
...
- *                                                                            *
- *             This software is distributed under the terms of the            *
- *                 GNU General Public Licence (GPL) version 3,                *
- *                    copied verbatim in the file "LICENSE".                  *
- *                                                                            *
- * In applying this license GSI does not waive the privileges and immunities  *
- * granted to it by virtue of its status as an Intergovernmental Organization *
- * or submit itself to any jurisdiction.                                      *
- ******************************************************************************/
+/****************************************************************************
```
jose-luis-rs: The same question

jose-luis-rs (changes requested):
Please check the minor changes required for approval

Mrunmoy-Jena:
Thanks for pointing out, I have now restored the headers back to the original format.

jose-luis-rs:
@Mrunmoy-Jena, you may consider including the angular dispersion of the incoming beam, as it would also affect the CALIFA response

Mrunmoy-Jena:
@jose-luis-rs Thank you for the suggestion. In my opinion, the angular dispersion of incoming beam (which corresponds to then an x,y spread in vertex position) is accounted for already in a sense by introducing the vertex distribution. What is important for the CALIFA response function is to include an angular spread of the outgoing fragment and sample theta (of the gamma wrt to outgoing fragment) from it (instead of simply taking CALIFA cluster theta) which is then to be used for the proper Doppler correction. 
I have modified the R3BCALIFATestGenerator class accordingly, and will update the pull request with the modified files if you agree to this.

jose-luis-rs:
At the entrance of the LH2 target, the beam is characterized by the mean values of its positions and angles, as well as by the corresponding spatial and angular spreads (dispersion). So your point is correct :+1: 

Please don't forget to apply the clang-format.

Mrunmoy-Jena:
Okay, now I updated it and ensured clang-format is applied. Thanks for the useful suggestions !

jose-luis-rs (approved):

